PR #15, opened 2019-11-18
Performance measurement skeleton

semtexzv:


prototypes/go/app/listener/listener.go
```diff
@@ -43,6 +44,34 @@ func shutdown() {
 	}
 }
 
+func runMessage(benchmark *Benchmark, m kafka.Message, mutex *sync.Mutex) {
```
Josca: This is not correct go way. It's better to use channel than mutex here, as channel is like the first class citizen in Go. So create one loop reading from channel messages and writing to database and then let's create go-routines sending results to this channel.

semtexzv: Yes, but for the purposes of this prototype, Mutex works well enough.

PR #18, opened 2019-11-21
Added GraphQL API example

Josca:
SPM-11


semtexzv:
Do we want to keep this in master ? From what I've understood, this was supposed to just be an evaluation. 

Also, we'll have to see how this relates to UI & requirement to implement REST API.

Josca:
@semtexzv it's up to you. We can either leave it unmerged or merge it and then remove to have it at least in history.

PR #19, opened 2019-11-25
Added API mocks

Josca:
- SPM-15


manager/controllers/advisory_detail.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// @Summary Show me details an advisory by given advisory name
+// @Description Show me details an advisory by given advisory name
+// @Accept   json
+// @Produce  json
+// @Param    advisory_id    path    string   true "Advisory ID"
+// @Success 200 {object} AdvisoryDetailResponse
+// @Router /api/patch/v1/advisories/{advisory_id} [get]
+func AdvisoryDetailHandler(c *gin.Context) {
```
semtexzv: This method should return a struct, which contains the `attributes` `id` and `type` in a `data` field, so that the format is the same between `/systems/id` and `/advisories/id`

docs/docs.go
```diff
@@ -0,0 +1,582 @@
+// GENERATED BY THE COMMAND ABOVE; DO NOT EDIT
```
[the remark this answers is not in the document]
Josca: I don't expect that API methods will be updated to often. So to keep it in git is easier here.

manager/controllers/structures.go
```diff
@@ -0,0 +1,56 @@
+package controllers
+
+import "time"
+
+type Links struct {
+	First    string  `json:"first"`
+	Last     string  `json:"last"`
+	Next     string  `json:"next"`
```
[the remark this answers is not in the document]
Josca: Yes, but now it works for mock ups. We can change it when we need.

manager/controllers/advisory_detail.go
Josca: no, because attributes of systems differ to attributes of advisory

manager/controllers/systems.go
```diff
@@ -0,0 +1,67 @@
...
+	"time"
+)
+
+type SystemsResponse struct {
+	Data  []SystemItem `json:"data"`
+	Links Links  `json:"links"`
+	Meta  SystemsMeta  `json:"meta"`
+}
+
+type SystemsMeta struct {
```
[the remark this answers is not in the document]
Josca: Do you mean something like:
~~~go
type AdvisoryMeta struct {
	Meta       Meta
	Severity   *string `json:"severity"`
}
~~~
?

I think that converting to json response wouldn't be so straightforward. In that case I don't mind a little of redundancy. But maybe I am wrong. Could you write example please?

manager/controllers/advisory_detail.go
semtexzv: I don't mean the attributes, I mean the document structure. Both of them should be

```json
{
    "data": {
        "attributes" : {
        
        },
        "id" : "<id>",
        "type" : "<type>"
    }
}
```

Josca: Ok, I see it now, thanks and sorry.

PR #20, opened 2019-11-27
Add RHEL dockerfiles

semtexzv:
- Added Dockerfiles for running the app on RHEL 7 & 8.
- Added script to check Dockerfiles consistency.
- We're waiting for CentOS images to catch up in order to migrate database to CentOS & Rhel 8


Dockerfile.rhel8
```diff
@@ -0,0 +1,26 @@
+FROM registry.access.redhat.com/ubi8
```
Josca: I would rather name these rhel based ones call Dockerfile, as it's default and production variant. And I would add centos suffix to dockerfiles based on centos because there are only for development purposes.

scripts/check-dockerfiles.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
Josca: I would prefer to have only one Dockerfile (rhel based) and then script to generate CentOS based one if developer needs it. But I know that this is the tested way which is used in vmaas and engine.

Dockerfile.rhel8
```diff
@@ -0,0 +1,26 @@
...
+
+ADD go.mod  /go/src/app/
+ADD go.sum  /go/src/app/
+
+WORKDIR /go/src/app
+RUN go mod vendor
+
+ADD /base       /go/src/app/base
+ADD /manager    /go/src/app/manager
+ADD /listener   /go/src/app/listener
```
Josca: you should add docs folder, as checking scrpt is warning you

Josca:
Please remove ".hel8/7" suffixes in production dockerfiles as discussed.

semtexzv:
I'd rather keep it as is. If someone will try to run the project, and they will try to use rhel based dockerfiles, they will encounter errors because of missing subscriptions. This variant contains most  information

PR #23, opened 2019-12-02
Openapi clients

semtexzv:
- Adds code for handling the 3scale identity strings
- Removes old `hosts` handling code & benchmarks
- Adds scripts to generate OpenApi clients into the _generated file

Goal of this PR is not to finish the whole evaluation flow, but to add OpenAPI generated clients and make listener able to call inventory and VMaaS.


scripts/generate-clients.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
Josca: It would be better to create separate repo(s) for generated vmaas and inventory clients and use it as libs. It avoids to storing scripts like this one here.
Btw. it does not work for me:
~~~
[main] INFO  o.o.codegen.DefaultGenerator - Model Facts not generated since it's a free-form object
[main] INFO  o.o.codegen.AbstractGenerator - writing file /local/_generated/cmsfr/inventory/model_bulk_host_out.go
Exception in thread "main" java.lang.RuntimeException: Could not generate model 'BulkHostOut'
	at org.openapitools.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:517)
	at org.openapitools.codegen.DefaultGenerator.generate(DefaultGenerator.java:923)
...
[main] INFO  o.o.codegen.AbstractGenerator - writing file /local/_generated/cmsfr/vmaas/model_cves_request.go
Exception in thread "main" java.lang.RuntimeException: Could not generate model 'CvesRequest'
~~~

Dockerfile.centos
```diff
@@ -4,8 +4,9 @@ RUN yum module -y install go-toolset postgresql && yum -y install git
 ENV GOPATH=/go
 ENV GO111MODULE=on
 
-ADD go.mod  /go/src/app/
-ADD go.sum  /go/src/app/
+ADD /_generated /go/src/app/_generated
```
Josca: As `_generated` folder is not included in repo after pull, it breaks many things that worked till now (docker-compose, go tests, docker build). When generating is not working then nothing is working.

base/utils/identity.go
```diff
@@ -0,0 +1,42 @@
...
+	// Additional information, we don't parse this
+	Internal      map[string]interface{}
+}
+
+type Identity struct {
+	Entitlements map[string]Entitlement `json:"entitlements"`
+	Identity     IdentityDetail         `json:"identity"`
+}
+
+func ParseIdentity(identityString string) (ident *Identity, err error) {
```
Josca: Named return values are not used here. Let's use types only.

docker-compose.yml
```diff
@@ -3,7 +3,6 @@ version: '3'
 services:
   
   db:
-    container_name: db
```
Josca: Thanks :smile:

scripts/generate-clients.sh
semtexzv: Moved the clients temporarily to https://github.com/semtexzv/patchman-clients, until we have repo in the RedhatInsights org. 

The error seems to be caused by user access rights to the directory. I'm running under docker, so its running with root .

base/utils/identity.go
semtexzv: Fixed

docker-compose.yml
semtexzv: You're welcome :)

semtexzv:
Moved the clients to separate repo. After we have created a dedicated repo in the RedHatIsights org, I'll move the clients there.

Dockerfile.centos
semtexzv: Fixed with the dedicated repository for clients.

listener/listener.go
```diff
@@ -37,43 +46,106 @@ func configure() {
...
+	config.Debug = true
+	config.BasePath = inventoryAddr + INVENTORY_API_PREFIX
+
+	inventoryClient = inventory.NewAPIClient(config)
+
+	cfg := vmaas.NewConfiguration()
+	cfg.BasePath = utils.GetenvOrFail("VMAAS_ADDRESS") + VMAAS_API_PREFIX
+	cfg.Debug = true
+
+	vmaasClient = vmaas.NewAPIClient(cfg)
```
Josca: Could we replace `base/vmaas_client` with that?

base/utils/identity_test.go
```diff
@@ -0,0 +1,27 @@
...
+		"ansible":          {Entitled: true},
+		"openshift":        {Entitled: true},
+	}
+
+	assert.Equal(t, nil, err)
+	assert.Equal(t, entitlements, identity.Entitlements)
+	assert.Equal(t, "6089719", identity.Identity.AccountNumber)
+
+	assert.Equal(t, true, identity.IsSmartEntitled())
+
```
Josca: empty. line

listener/listener.go
```diff
@@ -37,43 +46,106 @@ func configure() {
...
+	err := reader.Close()
 	if err != nil {
 		utils.Log("err", err.Error()).Error("unable to shutdown Kafka reader")
 	}
-
 }
 
 func baseListener(reader *kafka.Reader, handler func(message kafka.Message)) {
+	defer shutdown(reader)
+
```
Josca: Quite a lot of logic. How can I test whether it works?

listener/listener.go
```diff
@@ -37,43 +46,106 @@ func configure() {
...
+	var event PlatformEvent
+	utils.Log("msg", string(m.Value)).Info("Msg received")
+
+	err := json.Unmarshal(m.Value, &event)
+	if err != nil {
+		utils.Log("err", err.Error()).Error("Could not deserialize host event")
+		return
+	}
+	// We need the b64 identity in order to call the inventory
+	if event.B64Identity != nil && event.Url != nil {
```
Josca: Better to use standard go style here. That means check invalid states and handle it if needed. And keep expected routine on the left.
~~~go
step1()
if check1(): {
 log error; return
}
step2()
if check2(): {
 log error; return
}
step3()
...
~~~

listener/listener.go
```diff
@@ -37,43 +46,106 @@ func configure() {
...
+		if err != nil {
+			utils.Log("err", err.Error()).Error("Could not parse identity")
+			return
 		}
-		handler(m)
 
+
+		if identity.IsSmartEntitled() {
+			utils.Log("url", *event.Url).Error("Downloading system profile")
+			downloadSystemProfile(event.Id, *event.B64Identity)
```
Josca: This is our goal, so it should be in the end of method. Not nested in ifs.

listener/listener.go
```diff
@@ -37,43 +46,106 @@ func configure() {
...
-func logHandler(m kafka.Message) {
-	utils.Log().Info("Received message [", m.Topic,"] ", string(m.Value))
+func downloadSystemProfile(hostId string, identity string) {
+	utils.Log("hostId", hostId, "identity", identity).Error("Downloading system profile")
+
+	// Create new context, which has the apikey value set. This value is then used as a value for `x-rh-identity`
+	ctx := context.WithValue(context.Background(), inventory.ContextAPIKey, inventory.APIKey{Prefix: "", Key: identity})
+
+	data, res, err := inventoryClient.HostsApi.ApiHostGetHostSystemProfileById(ctx, []string{hostId}, nil)
+
```
Josca: emp. line

listener/listener.go
```diff
@@ -88,12 +160,11 @@ func RunListener() {
...
 	configure()
-	defer shutdown()
 
-	go baseListener(uploadReader, logHandler)
+	go baseListener(uploadReader, uploadHandler)
 	go baseListener(eventsReader, logHandler)
 
 	// Just block. Any error will panic and kill the process.
-	<- make(chan bool)
+	<-make(chan bool)
```
Josca: Could we replace it just second calling baseListener without go?

listener/listener.go
semtexzv: Well, that's the goal.

base/utils/identity_test.go
semtexzv: Which one ? We're gonna need to setup gofmt config and format code after each commit,

listener/listener.go
semtexzv: Fixed

listener/listener.go
semtexzv: Fixed.

base/utils/identity_test.go
Josca: Both or at least last empty line in function. I see no reason to have it there.

listener/listener.go
Josca: Ok, so could you remove base/vmaas_client package. Maybe you can use vmaas mock for testing your client, if you want.

listener/listener.go
```diff
@@ -37,43 +46,105 @@ func configure() {
...
-		handler(m)
-
+		// Spawn handler, not blocking the receiving goroutine
+		go handler(m)
 	}
 }
 
 func logHandler(m kafka.Message) {
-	utils.Log().Info("Received message [", m.Topic,"] ", string(m.Value))
+	utils.Log().Info("Received message [", m.Topic, "] ", string(m.Value))
```
Josca: better to use Log("topic", m.Topic) here

go.mod
```diff
@@ -4,37 +4,50 @@ go 1.11
 
 require (
 	github.com/DataDog/zstd v1.4.4 // indirect
+	github.com/RedHatInsights/patchman-clients/inventory v0.0.0-20191205080343-d3fd25e923bf
```
Josca: I would rather name it `go-api-clients` as these are not clients to patchman.

listener/listener.go
```diff
@@ -37,43 +46,105 @@ func configure() {
...
-		handler(m)
-
+		// Spawn handler, not blocking the receiving goroutine
+		go handler(m)
 	}
 }
 
 func logHandler(m kafka.Message) {
-	utils.Log().Info("Received message [", m.Topic,"] ", string(m.Value))
+	utils.Log("topic", m.Topic).Info("Received message ", string(m.Value))
```
Josca: Please use `Log("topic", m.Topic, "msg", string(m.Value))` instead.

PR #28, opened 2019-12-04
Add vmaas_sync component

semtexzv:
- Create new component
- Add openshift templates
- Client connecting to VMaaS websocket component


vmaas_sync/vmaas_sync.go
```diff
@@ -0,0 +1,54 @@
...
+		utils.Log("err", err.Error()).Error()
+		panic(err)
+	}
+}
+
+func runWebsocket(conn *websocket.Conn, handler func(data []byte, conn *websocket.Conn)) {
+	defer conn.Close()
+
+	for {
+		typ, msg, err := conn.ReadMessage()
```
jdobes: service also needs to write to websocket `subscribe-listener` message to receive notifications

see https://github.com/RedHatInsights/vulnerability-engine/blob/master/vmaas_sync/vmaas_sync.py#L299

semtexzv: Thanks, fixed.

vmaas_sync/vmaas_sync.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/gorilla/websocket"
+	ginprometheus "github.com/zsais/go-gin-prometheus"
+)
+
+func runMetrics() {
+	// create web app
+	app := gin.New()
+
+	prometheus := ginprometheus.NewPrometheus("gin")
```
Josca: Use `patchman_engine` metric prefix instead of `gin`, please.

Josca: Actually, there is middleware created to setup Prometheus metrics, please use it.

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/gin-contrib/gzip v0.0.1
 	github.com/gin-gonic/gin v1.4.0
 	github.com/go-sql-driver/mysql v1.4.1
+	github.com/gorilla/websocket v1.4.1 // indirect
```
Josca: This is used directly in vmaas_sync.go so remove `// indirect`, please. Btw., use `go mod tidy` command to clean up deps.

docker-compose.yml
```diff
@@ -52,13 +52,28 @@ services:
...
+    env_file:
+      - ./conf/common.env
+      - ./conf/vmaas_sync.env
+    command: ./vmaas_sync/entrypoint.sh
+    ports:
+      # Reserving 8082 for evaluator :)
+      - 8083:8083
+    depends_on:
+      - db
+      - platform
```
Josca: Component starts on `docker-compose up`, but how can I test that it's working? Could you add any testing scripts and description to README, please? I can not access :8083/metrics endpoint which should be exposed. Does it work for you?

vmaas_sync/vmaas_sync.go
```diff
@@ -0,0 +1,60 @@
...
+		panic(err)
+	}
+
+	for {
+		typ, msg, err := conn.ReadMessage()
+		if err != nil {
+			utils.Log("err", err.Error()).Fatal("Failed to retrive VMaaS websocket message")
+			panic(err)
+		}
+		if typ == websocket.BinaryMessage || typ == websocket.TextMessage {
```
Josca: Is message receiving info propagated into prom. metrics? How can I test it?

vmaas_sync/vmaas_sync.go
semtexzv: Fixed.

go.mod
semtexzv: Fixed

vmaas_sync/vmaas_sync.go
semtexzv: Added a simple prometheus metric

docker-compose.yml
semtexzv: There was issue with connecting 2 local docker-compose networks. Should be fixed now.

openshift/deploy/vmaas_sync.yml
```diff
@@ -0,0 +1,95 @@
+apiVersion: v1
```
beav: should this live here or in e2e-deploy?

semtexzv: Eventually. We're using ocdeployer, and keeping up with the conventions set in `e2e-deploy` but for the time being we're keeping the templates here. Once we're comfortable with the set of components and we need to integrate into the platform deployment process, we can very simply move them.

.env
```diff
@@ -0,0 +1 @@
+COMPOSE_PROJECT_NAME=vulnerability
```
Josca: vulnerability?

Dockerfile.centos
```diff
@@ -13,6 +13,7 @@ RUN go mod vendor
 ADD /base       /go/src/app/base
 ADD /manager    /go/src/app/manager
 ADD /listener   /go/src/app/listener
+ADD /vmaas_sync /go/src/app/vmaas_sync
```
Josca: Need to be added also to Dockerfile.test. Please, try always `docker-compose -f docker-compose.test.yml up --build test` before requesting review.

docker-compose.yml
```diff
@@ -7,14 +7,14 @@ services:
       context: .
       dockerfile: database/Dockerfile.centos
     ports:
-      - 5432:5432
+      - 15432:5432
```
Josca: Why the port is changing?

semtexzv: Because when we're running with VMaaS, it uses the 5432 port, this remaps the port only on host.

Dockerfile.centos
semtexzv: Fixed.

.env
semtexzv: For development, we need to connect to local VMaaS. We can only do this by running in a single project.

PR #29, opened 2019-12-05
List systems from db and return via API

Josca:
SPM-27

### DIFF
--- a/manager/controllers/paging.go
+++ b/manager/controllers/paging.go
@@ -1,0 +1,60 @@
+package controllers
+
+import "fmt"
+
+func CreateLinks(path string, offset, limit, total int, otherParams string) Links {
+	pager := pager{path, offset, limit, total, otherParams}
+	links := Links{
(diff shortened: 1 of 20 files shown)


manager/controllers/paging.go
```diff
@@ -0,0 +1,66 @@
+package controllers
+
+import "fmt"
+
+func CreateLinks(path string, offset, limit, total int, dataFormat string, showAll bool) Links {
```
semtexzv: What's the reason for show_all parameter ?

Josca: to show all systems or only for current account

jiridostal: It should allow you to show all Advisories, not only those that are affecting your environment. Currently it's not a requirement so we can remove it.

Josca:
@jiridostal @semtexzv I've removed `show_all` param and added generic `otherParams` param to enable inserting of additional query params co `CreateLinks` method.

PR #32, opened 2019-12-09
Add Batch insert + On conflict do update

semtexzv:


base/database/batch.go
```diff
@@ -0,0 +1,233 @@
...
+// nolint: gochecknoglobals
+var bulkNow time.Time
+
+func BulkInsert(db *gorm.DB, objects []interface{}) error {
+	return bulkExec(db, objects)
+}
+
+// BulkInsertChunk will split the objects passed into the passed chunk size. A
+// slice of errors will be returned (if any).
+func BulkInsertChunk(db *gorm.DB, objects []interface{}, chunkSize int) []error {
```
Josca: This is never used and tested.

base/database/batch.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return err
+	}
+
+	// No scope and no error means nothing to do
+	if scope == nil {
+		return nil
+	}
+
+	return db.Exec(scope.SQL, scope.SQLVars...).Error
```
Josca: This uses `INSERT INTO "test_tables" ("created_at", "deleted_at", "email", "name", "updated_at") VALUES (?, ?, ?, ?, ?), (?, ?, ?, ?, ?), (?, ?, ?, ?, ?) ` in tests. So it adds some extra fields we are not going to use.

base/database/dbtest.go
```diff
@@ -0,0 +1,18 @@
+package database
+
+import "github.com/jinzhu/gorm"
+
+type TestTable struct {
+	Name  string
+	Email string
+	gorm.Model
```
Josca: We don't use `gorm.Model` and `AutoMigrate`, so we should use the same style for testing and using.

base/database/batch.go
```diff
@@ -0,0 +1,233 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/jinzhu/gorm"
+)
+
+// bulkNow holds a global now state and will be used for each records as
```
Josca: Is this code copied from anywhere? We should reference it.

base/database/batch.go
semtexzv: Added a test for this function.

base/database/dbtest.go
semtexzv: Removed `gorm.Model`, but keeping the `AutoMigrate` as a way to simply generate table in the test database, without requiring  inclusion of the test table into the production database or custom SQL.

base/database/batch.go
semtexzv: Fixed.

base/database/batch.go
semtexzv: Behavior is the same as normal `gorm.Create`. The fields were added by `gorm.Model`.

PR #34, opened 2019-12-10
Automatic generation of OpenAPI 3.0 spec

semtexzv:
Uses approach from #24  with generating 2.0 spec using swag, and then converting it to 3.0.


scripts/generate_docs.sh
```diff
@@ -1,3 +1,26 @@
...
+#PID=$(docker run -d -p 28080:8080 --name swagger-converter swaggerapi/swagger-converter:v1.0.2)
+
+# Wait for converter to be ready
+until curl $CONVERT_URL > /dev/null 2> /dev/null; do
+  sleep 2
+done
+
+
+# Perform conversion
+curl -X "POST" -H  "accept: application/json" -H  "Content-Type: application/json" -d @$DOCS_TMP_DIR/swagger.json $CONVERT_URL > docs/openapi.json
```
beav: can you add a step at the end to feed the json through jsonlint or similar so it's pretty printed?

semtexzv: Yes, thanks.

PR #36, opened 2019-12-12
Database refactoring

semtexzv:
- Rename all occurences of `errata` to `advisory`, to correspond with frontend and API nomenclature
- Change `advisory_type` names to correspond to the ones we receive from VMaaS
- Remove `s3_url` from hosts, since we don't download archives.

### DIFF
--- a/base/models/models.go
+++ b/base/models/models.go
@@ -12,93 +12,80 @@ func (RhAccount) TableName() string {
 }
 
 type SystemPlatform struct {
-	ID               int
-	InventoryID      string
-	RhAccountID      int
-	FirstReported    time.Time
-	S3Url            string
-	VmaasJson        string
-	JsonChecksum     string
-	LastUpdated      time.Time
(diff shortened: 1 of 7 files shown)


base/models/models.go
```diff
@@ -12,93 +12,80 @@ func (RhAccount) TableName() string {
...
-	RhAccountID      int
-	FirstReported    time.Time
-	S3Url            string
-	VmaasJson        string
-	JsonChecksum     string
-	LastUpdated      time.Time
-	UnchangedSince   time.Time
-	LastEvaluation   *time.Time
-	OptOut           bool
-	ErrataCountCache int
```
beav: it looks like ErrataCountCache was removed as well. Just wanted to confirm this was intentional and not missed as part of the rename.

semtexzv: Renamed to AdvisoryCountCache

beav: ah i see it now, thx

PR #38, opened 2019-12-17
Connected API to data in database

Josca:
- Rest of manager endpoints connected to database.
- It does not include filtering with rh-account (another ticket).
- Not complete counting of system advisories. It requires bigger db schema update (another ticket).

### DIFF
--- a/scripts/test_data.sql
+++ b/scripts/test_data.sql
@@ -1,5 +1,11 @@
-DELETE FROM system_repo; DELETE FROM system_platform; DELETE FROM rh_account; DELETE FROM repo;
+DELETE FROM system_advisories;
+DELETE FROM system_repo;
+DELETE FROM system_platform;
+DELETE FROM repo;
 DELETE FROM timestamp_kv;
+DELETE FROM advisory_account_data;
+DELETE FROM advisory_metadata;
+DELETE FROM rh_account;
 
 INSERT INTO rh_account (id, name) VALUES
 (0, '0'), (1, '1'), (2, '2'), (3, '3');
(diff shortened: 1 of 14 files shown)


scripts/test_data.sql
```diff
@@ -26,7 +32,38 @@ INSERT INTO system_platform (id, inventory_id, rh_account_id,  vmaas_json, json_
 (10, 'INV-10', 1, '{ "package_list": [ "kernel-2.6.32-696.20.1.el6.x86_64" ]}', '1', '2018-09-22 12:00:00-04', '2018-09-22 12:00:00-04'),
 (11, 'INV-11', 2, '{ "package_list": [ "kernel-2.6.32-696.20.1.el6.x86_64" ]}', '1', '2018-09-22 12:00:00-04', '2018-09-22 12:00:00-04');
 
--- TODO add advasories_metadata, system_advasories, adv. account metadata
+INSERT INTO advisory_metadata (id, name, description, synopsis, summary, solution, advisory_type_id,
+                               public_date, modified_date, url) VALUES
+(1, 'RH-1', 'adv-1-des', 'adv-1-syn', 'adv-1-sum', 'adv-1-sol', 1, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url1'),
```
beav: can you make the test data more lifelike? for example, maybe "advisory 1 summary blah blah blah".

beav:
This works for me and tests pass

I added one comment but ACK otherwise.

scripts/test_data.sql
Josca: If I extend testing data it will extend whole file and make it less clear. I would like to keep it in this simple form. Also it's useful for tests because each value is easy to predict. Of course it's no problem to change it if you really really want :smile:

beav: I'm ok with leaving it like this. It will have real data soon enough anyway :)

PR #39, opened 2019-12-17
Save hosts in listener

semtexzv:
Moves code which processes uploads to separate file.  Implements process to get system_profile from inventory, create VMaaS request, And store both RHAccount and SystemProfile into the database.

Tests for parts which do not require external APIs are implemented.

### DIFF
--- a/base/models/models.go
+++ b/base/models/models.go
@@ -15,11 +15,12 @@ type SystemPlatform struct {
 	ID                    int
 	InventoryID           string
 	RhAccountID           int
-	FirstReported         time.Time
+	// All times need to be stored as pointers, since they are set to 0000-00-00 00:00 by gorm if not present
+	FirstReported         *time.Time
 	VmaasJSON             string
 	JsonChecksum          string
-	LastUpdated           time.Time
-	UnchangedSince        time.Time
+	LastUpdated           *time.Time
+	UnchangedSince        *time.Time
 	LastEvaluation        *time.Time
 	OptOut                bool
 	AdvisoryCountCache    int
@@ -54,7 +55,7 @@ type SystemAdvisories struct {
 	ID            int
 	SystemID      int
 	AdvisoryID    int
-	FirstReported time.Time
+	FirstReported *time.Time
 	WhenPatched   *time.Time
 	StatusID      *int
 }
@@ -73,6 +74,16 @@ type AdvisoryAccountData struct {
 
 func (AdvisoryAccountData) TableName() string {
 	return "advisory_account_data"
+}
+
+type SystemAdvisoriesSlice []SystemAdvisories
+
+func (this SystemAdvisoriesSlice) ToInterfaceSlice() []interface{} {
+	res := make([]interface{}, len(this))
+	for i, v := range this {
+		res[i] = v
+	}
+	return res
 }
 
 type Repo struct {
--- a/listener/listener.go
+++ b/listener/listener.go
@@ -1,18 +1,13 @@
 package listener
 
 import (
-	"app/base/database"
-	"app/base/models"
 	"app/base/utils"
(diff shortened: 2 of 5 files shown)


base/models/models.go
```diff
@@ -69,6 +70,15 @@ type AdvisoryAccountData struct {
 func (AdvisoryAccountData) TableName() string {
 	return "advisory_account_data"
 }
+type SystemAdvisoriesSlice []SystemAdvisories
+
+func (this SystemAdvisoriesSlice) ToInterfaceSlice() []interface{} {
+	var res []interface{}
```
Josca: As we know source array size, could we allocate result array in the beginning?

Josca: The same optimization could be done in the `OnConflictUpdateMulti` method.

listener/upload.go
```diff
@@ -0,0 +1,153 @@
...
+	hash := sha256.New()
+	hash.Write(updatesReqJSON)
+	jsonChecksum := hex.EncodeToString(hash.Sum([]byte{}))
+
+	now := time.Now()
+
+	dbHost := models.SystemPlatform{
+		InventoryID:    inventoryId,
+		RhAccountID:    accountId,
+		FirstReported:  &now,
```
Josca: Is it sure, that this is the first upload?

listener/upload.go
```diff
@@ -0,0 +1,153 @@
...
+	"encoding/json"
+	"github.com/RedHatInsights/patchman-clients/inventory"
+	"github.com/RedHatInsights/patchman-clients/vmaas"
+	"github.com/antihax/optional"
+	"github.com/segmentio/kafka-go"
+	"time"
+)
+
+func uploadHandler(m kafka.Message) {
+	var event PlatformEvent
```
Josca: How can I test it? Is there any script or guide to run all workflow using docker-compose?

base/models/models.go
```diff
@@ -51,7 +52,7 @@ type SystemAdvisories struct {
 	ID            int
 	SystemID      int
 	AdvisoryID    int
-	FirstReported time.Time
+	FirstReported *time.Time
```
Josca: This field corresponds to `first_reported` column in db which is NOT NULL, so it doesn't have to be pointer.

listener/listener.go
```diff
@@ -46,19 +44,21 @@ func configure() {
...
 	eventsReader = kafka.NewReader(eventsConfig)
 
+	traceApi := utils.GetenvOrFail("LOG_LEVEL") == "trace"
+
 	inventoryAddr := utils.GetenvOrFail("INVENTORY_ADDRESS")
 
-	config := inventory.NewConfiguration()
-	config.Debug = true
-	config.BasePath = inventoryAddr + INVENTORY_API_PREFIX
+	invConfig := inventory.NewConfiguration()
```
Josca: Yes, these names are better. But I wouldn't use shortcut here, but name it simply `inventoryConfig`.

base/models/models.go
semtexzv: If it isn't null its set to `0000-00- .....` By gorm. This is wrong. The field is automatically set to current timestamp by a database trigger if null is provided.

base/models/models.go
semtexzv: Fixed, thanks

listener/upload.go
semtexzv: This field is set by database trigger. Fixed.

listener/listener.go
semtexzv: Fixed.

listener/upload.go
semtexzv: Not yet, we'll need to either extend Unit tests to run with platform, or introduce separate tools for testing whole workflows ...

PR #43, opened 2020-01-03
Upload tests

Josca:
- Evaluation moved to separate file and method.
- Added kafka message parsing test.
- Added upload handler test.

### DIFF
--- a/listener/upload_test.go
+++ b/listener/upload_test.go
@@ -6,6 +6,7 @@ import (
 	"app/base/models"
 	"app/base/utils"
 	"github.com/RedHatInsights/patchman-clients/vmaas"
+	"github.com/segmentio/kafka-go"
 	"github.com/stretchr/testify/assert"
 	"testing"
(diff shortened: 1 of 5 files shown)


listener/upload_test.go
```diff
@@ -74,3 +75,32 @@ func TestUpdateSystemPlatform(t *testing.T) {
...
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+	configure()
+	deleteData(t)
+
+	msg := createTestingUploadKafkaMsg()
+	uploadHandler(msg)
+
+	var system models.SystemPlatform
+	assert.Nil(t, database.Db.Where("inventory_id = ?", "TEST-00000").Find(&system).Error)
```
semtexzv: Could you please add more assertions, for user account, and for different attributes of SystemPlatform like last_seen being close to now, and last_evaluated being null ?

PR #45, opened 2020-01-06
Run tests on Travis

mkoura:
Run unit tests on Travis. I don't have rights to enable the repo on Travis, can some of you please do it? - https://travis-ci.org/RedHatInsights/patchman-engine/

Also adds upload to codecov.io (only when running on Travis).

Test run on https://travis-ci.org/mkoura/patchman-engine/builds/633258180

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -28,7 +28,19 @@ services:
       dockerfile: Dockerfile.test
     env_file:
       - ./conf/test.env
+    environment:
+      - CI
+      - TRAVIS
+      - SHIPPABLE
+      - TRAVIS_BRANCH
+      - TRAVIS_COMMIT
+      - TRAVIS_JOB_NUMBER
+      - TRAVIS_PULL_REQUEST
+      - TRAVIS_JOB_ID
+      - TRAVIS_REPO_SLUG
+      - TRAVIS_TAG
+      - TRAVIS_OS_NAME
     depends_on:
       - db
       - platform
-    command: ./scripts/go_test_db.sh
+    command: ./scripts/go_test_on_ci.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+git:
+  depth: false
+branches:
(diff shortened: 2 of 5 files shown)


Josca:
@mkoura ok, I've added repo to Travis. Now there are some conflicts, could you resolve it, please? Another thing - I've added dependency on platform container in the last PR. Seed docker-compose.test.yml. Could you add it also to travis config?

beav:
@mkoura is something else needed to get travis to run on this PR?

mkoura:
@beav I still can see

>  This is not an active repository
>
> You don't have sufficient rights to enable this repo on Travis.
> Please contact the admin to enable it or to receive admin rights yourself. 

on https://travis-ci.org/RedHatInsights/patchman-engine/

I guess that's the reason why travis hasn't tested this PR yet.

Josca:
@mkoura I see the same message about rights. @semtexzv could you help us to resolve this or could you increase our rights over this repo?

semtexzv:
Travis was enabled.

Josca:
@mkoura Travis was enabled, but [first build failed](https://travis-ci.org/RedHatInsights/patchman-engine/builds/633660537). How can we test your PR with Travis config? Could you do another push to your branch?

docker-compose.test.yml
```diff
@@ -28,7 +28,19 @@ services:
       dockerfile: Dockerfile.test
     env_file:
       - ./conf/test.env
+    environment:
```
Josca: Is it possible to move these vars to new env file `./conf/travis.env`? I think it's better to keep this file as brief as possible.

Josca: Actually, why do we need all these vars when they are not set to any value?

mkoura: That means these variables and their values are passed from environment. Meaning when running on travis, these will be set. I don't think variables without values can be specified in `./conf/travis.env`.

semtexzv:
:fireworks:

PR #48, opened 2020-01-07
Implement authentication & account separation for hosts

semtexzv:
This PR adds middleware for parsing `x-rh-identity` header, and SQL clauses to restrict 
systems returned based on account_id provided in the identity header.


Josca:
@semtexzv please, solve these conflicts

manager/controllers/systems_auth_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/", nil)
+	initRouterWithAccount(SystemsListHandler, "/", acc).ServeHTTP(w, req)
+	assert.Equal(t, 200, w.Code)
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+	// data
+	assert.Equal(t, count, len(output.Data))
+}
+func TestMissingAccount(t *testing.T) {
```
Josca: Please, put empty line here.

dev/scripts/advisories_list.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 
-curl -v -XGET http://localhost:8080/api/patch/v1/advisories | python -m json.tool
+source "$(dirname $(realpath "$0"))/env.sh"
```
Josca: Instead of this I would add DEV_MODE env var to manager and when it's active put testing x-rh-identity string using some "middlewares.PutTestingIdentity()" to each request. It enables also calling from browser which is now broken (I guess). Then we can also remove this IDENTITY env var setting and passing to each request.

manager/routes/routes.go
```diff
@@ -13,6 +14,7 @@ func Init(app *gin.Engine) {
 }
 
 func InitAPI(group *gin.RouterGroup) {
+	group.Use(middlewares.Authenticator())
```
Josca: Yes, here it the place I would conditionally put `middlewares.PutTestingIdentity()`.

manager/middlewares/authentication.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"app/base/utils"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+const KEY_ACCOUNT = "account"
+
+
+type ErrorResponse struct {
```
Josca: This struct already exists in controllers.structures. So what to move into base.utils.gin and use it in both places?

dev/scripts/advisories_list.sh
semtexzv: This is how its done in vulnerabilities. I'd like to be able to query the DB as different users, and not introduce different behavior into the base manager code. For unit tests, we use MockAuthenticator.

manager/controllers/systems_auth_test.go
semtexzv: Fixed

manager/middlewares/authentication.go
semtexzv: Thanks, fixed.

PR #58, opened 2020-01-14
Move to v3/updates call, add more info to vmaas request

semtexzv:
VMaaS call for `/v3/updates` was implemented, this PR changes evaluation behaviour to use this endpoint, and adds modules + repos to VMaaS request


beav:
changes look good, will ack after automation runs and is green

PR #60, opened 2020-01-14
Added CloudWatch logging

Josca:


beav:
```
manager_1     | time="2020-01-15T13:22:32Z" level=info msg="config for aws CloudWatch not loaded"
```

This is about the best we can do for testing outside of CI. ACK

PR #61, opened 2020-01-15
Increase kafka timeout

semtexzv:


beav:
was it timing out before?

semtexzv:
Yes, we have timeouts when we receive large upload batches, this change is to see whether it is just a matter of kafka being overloaded or other issue.

PR #62, opened 2020-01-15
Batch insert improvements

semtexzv:
Implements 2 changes to batch insert code.
- No longer require `[]interface{}`, allowing user to pass in any slice as an input
- Replicate `Create` behavior of scanning inserted values back into the input slice, reducing the need to perform second query to load just inserted values


Dockerfile.test
```diff
@@ -1,6 +1,6 @@
 FROM centos:8
 
-RUN yum module -y install go-toolset postgresql && yum -y install git
+RUN yum module -y install go-toolset postgresql && yum -y install git diffutils
```
Josca: Where is it used? Isn't it because of database migrations? So it should be part of its PR.

base/database/batch.go
```diff
@@ -157,8 +186,9 @@ func insertFunc(scope *gorm.Scope, columnNames, groups []string) {
 		extraOptions = fmt.Sprintf(" %s", insertOption)
 	}
 
-	scope.Raw(fmt.Sprintf( //nolint:gosec
-		"INSERT INTO %s (%s) VALUES %s %s",
+	// nolint: gosec
```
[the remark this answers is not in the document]
semtexzv: No, since we're not binding just values, but also column names. There is quoting of the values, but this is potential SQL injection vector if we ever accept unrestricted user input for column names.

Dockerfile.test
semtexzv: When golint fails, it tries to render output with diff.

PR #64, opened 2020-01-15
Added golangci-lint checker with default settings

Josca:
- added automatic checking
- fixed lint hints

### DIFF
--- a/base/database/setup.go
+++ b/base/database/setup.go
@@ -83,16 +83,6 @@ func loadEnvPostgreSQLConfig() *PostgreSQLConfig {
 	return &config
 }
 
-// create config string from additional timeout params
-func timeoutParams(dbConfig *PostgreSQLConfig) string {
-	return fmt.Sprintf(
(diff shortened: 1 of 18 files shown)


base/database/setup.go
```diff
@@ -83,16 +83,6 @@ func loadEnvPostgreSQLConfig() *PostgreSQLConfig {
 	return &config
 }
 
-// create config string from additional timeout params
-func timeoutParams(dbConfig *PostgreSQLConfig) string {
```
beav: was this unused?

beav:
@Josca can you check if the linting was executed in the PR tests? i didn't see it but its very possible I missed it when scrolling through.

base/database/setup.go
Josca: yes

Josca:
Yes, it was only visible when code analysis failed. It was not ideal. I've added simple [message](https://github.com/RedHatInsights/patchman-engine/pull/64/checks?check_run_id=392736900#step:3:2173) after analysis finishing.

PR #65, opened 2020-01-16
Database migrations

semtexzv:
Database migrations inspired by work on vulnerability-engine, utilizing go-migrate.

- Split schema into 4 separate migrations
- Added test which verifies final schema & migrations are compatible
- Changed `SERIAL` columns to `GENERATED BY DEFAULT AS IDENTITY` to make postgres handling of sequences internal.


database/migrations/002_base_schema.up.sql
```diff
@@ -0,0 +1,244 @@
...
+
+-- system_platform
+CREATE TABLE IF NOT EXISTS system_platform
+(
+    id                       INT GENERATED BY DEFAULT AS IDENTITY,
+    inventory_id             TEXT                     NOT NULL,
+    CHECK (NOT empty(inventory_id)),
+    rh_account_id            INT                      NOT NULL,
+    first_reported           TIMESTAMP WITH TIME ZONE NOT NULL,
+    vmaas_json               TEXT,
```
beav: If this is guaranteed to be json, would `JSONB` work better than `TEXT`?

beav:
I get this error:

```
db_1          | => sourcing /usr/share/container-scripts/postgresql/start/init_schema.sh ...
db_1          | Already initialized - Migrating the database
db_1          | error: Dirty database version 2. Fix and force version.
```

What is the best way to clear the db?

semtexzv:
You need to delete old volumes. `docker-compose down -v` should do it.

beav:
i had one question about TEXT vs JSONB but looks good

database/migrations/002_base_schema.up.sql
semtexzv: There were some discussions about this. But I think then the checksum would not work, since JSONB might reformat the query. 

So we either remove the checksum or utilize JSONB

scripts/go_test.sh
```diff
@@ -2,6 +2,17 @@
...
+  # Run go test and colorize output (PASS - green, FAIL - red).
+  # Set "-p 1" to run test sequentially to avoid parallel changes in testing database.
+  go test -v -p 1 -coverprofile=coverage.txt -covermode=atomic $1 | sed ''/PASS/s//$(printf "\033[32mPASS\033[0m")/'' | sed ''/FAIL/s//$(printf "\033[31mFAIL\033[0m")/''
+}
+
+# Run normal tests
+run_tests "$TEST_DIRS"
+
+# Run destructive schema compatiblity test
+run_tests app/database
```
Josca: So will this remove testing data and schema? I use testing database after running tests in docker-compose. Testing database initialization is the useful side effect of tests running. Also I am not sure if it's good to send coverage report twice. It's currently broken. I reported it to @mkoura.

semtexzv: I didn't think of that. Yes it does.

I think we need to have this test. At least in the CI pipeline.

Josca: Could you modify this test so it will leave the data unchanged? You can process 1) drop data 2) down pipeline - 3) check - 4) up pipeline - 5) check and finally apply test_data.sql again. It allows you to include this test among the others. So you will send coverage report only once.

Josca:
@semtexzv could you resolve conflicts and push update? Now it should send coverage report as it was fixed by @mkoura.

database/Dockerfile.centos
```diff
@@ -1,5 +1,17 @@
+FROM byrnedo/alpine-curl
```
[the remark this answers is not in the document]
semtexzv: Fixed

scripts/go_test.sh
semtexzv: Thanks, Schema test now runs before unit tests.

PR #67, opened 2020-01-20
Add common sort fields to advisories & systems endpoints

semtexzv:


manager/controllers/advisories.go
```diff
@@ -9,6 +9,8 @@ import (
 	"net/http"
 )
 
+var AdvisoriesSortFields = []string{"name", "description", "synopsis", "summary", "solution", "public_date"}
```
beav: Is the sort on text fields needed? We had issues with Katello in this area in the past, as the number of errata grows.

manager/controllers/utils.go
```diff
@@ -25,7 +25,7 @@ func LogAndRespNotFound(c *gin.Context, err error, respMsg string) {
 	c.JSON(http.StatusNotFound, utils.ErrorResponse{Error: respMsg})
 }
 
-func ApplySort(c *gin.Context, tx *gorm.DB, allowedFields ...string) (*gorm.DB, error) {
+func ApplySort(c *gin.Context, tx *gorm.DB, tablePrefix string, allowedFields ...string) (*gorm.DB, error) {
```
Josca: I would remove this tablePrefix feature. Let's suppose no prefix. We can solve it using AS keyword on query level. This solution does not allow usage of columns from two tables (two different prefixes) in case of join anyway. What do you mean @beav ?

manager/controllers/advisories.go
Josca: As @mkoura mentioned before, we should also update openapi spec when adding filtering.

manager/controllers/utils.go
semtexzv: That would be better solution. But it does not work when there are 2 tables in a join clause which both have sort attribute.

semtexzv:
Added explicit selects & openapi docs

PR #69, opened 2020-01-22
replace Github Actions with Travis

mkoura:


Dockerfile.centos
```diff
@@ -1,6 +1,8 @@
 FROM centos:8
 
-RUN yum module -y install go-toolset postgresql && yum -y install git
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial && \
```
mkoura: I needed to add implicit import of keys to all containers that are using centos, otherwise the build would fail like in https://travis-ci.org/RedHatInsights/patchman-engine/builds/640387466?utm_medium=notification&utm_source=github_status

Josca: ok

PR #70, opened 2020-01-22
Added metrics to monitor database items

Josca:
- systems stats
- advisories stats
- system advisories stats


semtexzv:
I don't thinks these metrics should be in manager. I'd move these VMaaS sync & focus on reporting metrics only about endpoints here.

Josca:
@semtexzv ok, I moved new metrics to be served from vmaas_sync. We already have monitoring of manager endpoints.

PR #72, opened 2020-01-22
Added received messages metric to listener

Josca:
- received messages metric
- evaluation duration metric
- updates count metric, etc

### DIFF
--- a/database/schema/create_schema.sql
+++ b/database/schema/create_schema.sql
@@ -583,6 +583,7 @@ CREATE INDEX ON advisory_metadata (advisory_type_id);
 
 GRANT SELECT, INSERT, UPDATE, DELETE ON advisory_metadata TO evaluator;
 GRANT SELECT, INSERT, UPDATE, DELETE ON advisory_metadata TO vmaas_sync;
+GRANT SELECT, INSERT, UPDATE, DELETE ON advisory_metadata TO listener;
 
 
 -- status table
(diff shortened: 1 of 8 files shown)


database/schema/create_schema.sql
```diff
@@ -587,6 +588,7 @@ CREATE TRIGGER system_advisories_set_first_reported BEFORE INSERT ON system_advi
   FOR EACH ROW EXECUTE PROCEDURE set_first_reported();
 
 GRANT SELECT, INSERT, UPDATE, DELETE ON system_advisories TO evaluator;
+GRANT SELECT, INSERT, UPDATE, DELETE ON system_advisories TO listener;
```
semtexzv: We'll have to remove this permissions once we separate evaluator from listener.

listener/upload.go
```diff
@@ -18,16 +18,25 @@ import (
 func uploadHandler(event PlatformEvent) {
 	if event.B64Identity == nil {
 		utils.Log().Error("Identity not provided")
+		messagesReceivedCnt.WithLabelValues("error-identity").Inc()
```
semtexzv: Please add prefix or separate metrics to differentiate between upload & delete messages.

database/schema/create_schema.sql
Josca: Yes, but it wouldn't be too difficult :smile:

listener/upload.go
Josca: ok, done

PR #73, opened 2020-01-23
Fixed error (500) response on not allowed filter

Josca:


manager/controllers/advisory_systems.go
```diff
@@ -61,7 +61,7 @@ func AdvisorySystemsListHandler(c *gin.Context) {
 
 	query, err = ApplySort(c, query, SystemsSortFields...)
 	if err != nil {
-		LogAndRespError(c, err, err.Error())
+		LogAndRespBadRequest(c, err, "sort application failed")
```
semtexzv: Could we add a test case for this ? Returning invalid request on invalid sort field.

Josca: Yes, I would create ticket to increase test coverage in the next sprint. Or do you really want to add it into this PR?

semtexzv: Why not, it's a short PR

Josca: ok, test added

PR #76, opened 2020-01-27
Make resource limits more reasonable

semtexzv:
- Listener does almost nothing, Current implementation even less so. but we need multiple pods
- Database runs on 300MB RAM right now, without any traffic


database/schema/init_schema.sh
```diff
@@ -12,9 +12,9 @@ if $PG_INITIALIZED; then
   psql -d ${POSTGRESQL_DATABASE} -f ${CONTAINER_SCRIPTS_PATH}/start/create_users.sql
   # Create schema from scratch
   psql -d ${POSTGRESQL_DATABASE} -f ${CONTAINER_SCRIPTS_PATH}/start/create_schema.sql
-else
-  echo "Already initialized - Migrating the database"
-  ${CONTAINER_SCRIPTS_PATH}/migrate.sh up
+#else
```
Josca: why do we comment that?

Josca: Discussed personally - db migration does not work currently. It should be fixed.

PR #78, opened 2020-01-28
Added new duration metrics

Josca:


evaluator/metrics.go
```diff
@@ -17,12 +17,12 @@ var (
 		Name:      "updates",
 	}, []string{"type"})
 
-	evaluationDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+	evaluationDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "patchman_engine",
 		Subsystem: "evaluator",
 		Name:      "evaluation_duration_milliseconds",
 		Buckets:   []float64{.00001, .0001, .001, .01, .1, 1, 10, 100, 1000},
```
semtexzv: We need the most resolution in the 1ms to ~10000ms range.

listener/delete.go
```diff
@@ -4,9 +4,13 @@ import (
 	"app/base/database"
 	"app/base/mqueue"
 	"app/base/utils"
+	"time"
 )
 
 func deleteHandler(event mqueue.PlatformEvent) {
+	tStart := time.Now()
+	defer messageHandlingDuration.WithLabelValues("delete").Observe(time.Since(tStart).Seconds())
```
semtexzv: There is EventDelete const.

listener/metrics.go
```diff
@@ -23,10 +23,16 @@ var (
 		Subsystem: "listener",
 		Name:      "kafka_message_received",
 	}, []string{"event", "type"})
+
+	messageHandlingDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "listener",
+		Name:      "kafka_message_handling_duration_seconds",
```
semtexzv: Missing buckets, I'd also unify the names between listener and evaluator, something like `handle_duration`

Josca: We can use default buckets here. This is the convention according to Prometheus docs to add duration_{unit} suffix. I wouldn't add "handle" word to evaluation duration metric name.

PR #80, opened 2020-01-29
Advisory severity

semtexzv:
Sync advisory severity from vmaas.
- After internal discussion interprets `None` as null. Provides only index in the frontend, since strings will always be translated.

### DIFF
--- a/scripts/test_data.sql
+++ b/scripts/test_data.sql
@@ -29,15 +29,15 @@ INSERT INTO system_platform (id, inventory_id, rh_account_id,  vmaas_json, json_
 (11, 'INV-11', 2, '{ "package_list": [ "kernel-2.6.32-696.20.1.el6.x86_64" ]}', '1', '2018-09-22 12:00:00-04', '2018-09-22 12:00:00-04');
 
 INSERT INTO advisory_metadata (id, name, description, synopsis, summary, solution, advisory_type_id,
-                               public_date, modified_date, url) VALUES
-(1, 'RH-1', 'adv-1-des', 'adv-1-syn', 'adv-1-sum', 'adv-1-sol', 1, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url1'),
-(2, 'RH-2', 'adv-2-des', 'adv-2-syn', 'adv-2-sum', 'adv-2-sol', 2, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url2'),
-(3, 'RH-3', 'adv-3-des', 'adv-3-syn', 'adv-3-sum', 'adv-3-sol', 3, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url3'),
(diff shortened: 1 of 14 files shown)


scripts/test_data.sql
```diff
@@ -29,15 +29,15 @@ INSERT INTO system_platform (id, inventory_id, rh_account_id,  vmaas_json, json_
...
-(2, 'RH-2', 'adv-2-des', 'adv-2-syn', 'adv-2-sum', 'adv-2-sol', 2, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url2'),
-(3, 'RH-3', 'adv-3-des', 'adv-3-syn', 'adv-3-sum', 'adv-3-sol', 3, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url3'),
-(4, 'RH-4', 'adv-4-des', 'adv-4-syn', 'adv-4-sum', 'adv-4-sol', 1, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url4'),
-(5, 'RH-5', 'adv-5-des', 'adv-5-syn', 'adv-5-sum', 'adv-5-sol', 2, '2016-09-22 12:00:00-05', '2017-09-22 12:00:00-05', 'url5'),
-(6, 'RH-6', 'adv-6-des', 'adv-6-syn', 'adv-6-sum', 'adv-6-sol', 3, '2016-09-22 12:00:00-06', '2017-09-22 12:00:00-06', 'url6'),
-(7, 'RH-7', 'adv-7-des', 'adv-7-syn', 'adv-7-sum', 'adv-7-sol', 1, '2017-09-22 12:00:00-07', '2017-09-22 12:00:00-07', 'url7'),
-(8, 'RH-8', 'adv-8-des', 'adv-8-syn', 'adv-8-sum', 'adv-8-sol', 2, '2016-09-22 12:00:00-08', '2018-09-22 12:00:00-08', 'url8');
+                               public_date, modified_date, url, severity_id) VALUES
+(1, 'RH-1', 'adv-1-des', 'adv-1-syn', 'adv-1-sum', 'adv-1-sol', 1, '2016-09-22 12:00:00-04', '2017-09-22 12:00:00-04', 'url1', NULL),
```
Josca: Isn't better to set not-null value for severity_id in testing data as it's closer to real data?

vmaas_sync/advisory_sync.go
```diff
@@ -60,14 +87,19 @@ func parseAdvisories(data map[string]vmaas.ErrataResponseErrataList) (models.Adv
...
 		}
+		var severityID *int
+		if id, has := severities[strings.ToLower(v.Severity)]; has {
+			severityID = &id
+		}
 
 		advisory := models.AdvisoryMetadata{
 			Name:           n,
-			AdvisoryTypeID: advisoryTypes[v.Type],
+			AdvisoryTypeID: advisoryTypes[strings.ToLower(v.Type)],
```
Josca: Why isn't it also checked as severityID?

semtexzv: It's always non-null & there is foreign key. If there is invalid input, it sets advisory type to 0, and fails to save the metadata.

scripts/test_data.sql
semtexzv: There are severities set for RHSAs in this list.

PR #82, opened 2020-01-29
Unified list meta field & limit, offset, sort handling

semtexzv:
- Unifies different `meta` fields.
- Handles limit, offset, sort in common function

### DIFF
--- a/manager/controllers/utils.go
+++ b/manager/controllers/utils.go
@@ -11,21 +11,24 @@ import (
 	"strings"
 )
 
+const InvalidOffsetMsg = "Invalid offset"
+
 func LogAndRespError(c *gin.Context, err error, respMsg string) {
 	utils.Log("err", err.Error()).Error(respMsg)
-	c.JSON(http.StatusInternalServerError, utils.ErrorResponse{Error: respMsg})
+	c.AbortWithStatusJSON(http.StatusInternalServerError, utils.ErrorResponse{Error: respMsg})
 }
 
 func LogAndRespBadRequest(c *gin.Context, err error, respMsg string) {
 	utils.Log("err", err.Error()).Warn(respMsg)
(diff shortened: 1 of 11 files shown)


manager/controllers/utils.go
```diff
@@ -10,43 +11,92 @@ import (
...
+
+	var total int
+	err = tx.Count(&total).Error
+	if err != nil {
+		LogAndRespError(c, err, "Database connection error")
+		return nil, nil, nil, err
+	}
+
+	if offset > total {
+		err = errors.New("Offset")
```
Josca: would add more descriptive message, e.g: "offset greater than total"

manager/controllers/utils.go
```diff
@@ -10,43 +11,92 @@ import (
...
-			return nil, errors.Errorf("Invalid sort field: %v", enteredField)
+			return nil, nil, errors.Errorf("Invalid sort field: %v", enteredField)
 		}
+		appliedFields = append(appliedFields, enteredField)
+	}
+	return tx, appliedFields, nil
+}
+
+// nolint:lll
+func ListCommon(tx *gorm.DB, c *gin.Context, allowedFields []string, path string) (*gorm.DB, *ListMeta, *Links, error) {
```
Josca: In case of so many return vals maybe it's better to use named return vals. Yes but here it's not so needed because of different types of return values.

manager/controllers/utils.go
```diff
@@ -10,43 +11,92 @@ import (
...
 			// We have not found any matches in allowed fields, return an error
-			return nil, errors.Errorf("Invalid sort field: %v", enteredField)
+			return nil, nil, errors.Errorf("Invalid sort field: %v", enteredField)
 		}
+		appliedFields = append(appliedFields, enteredField)
+	}
+	return tx, appliedFields, nil
+}
+
+// nolint:lll
```
Josca: :smile: cheeting?

PR #92, opened 2020-01-31
Replace docker with podman

Josca:


beav:
I tested this with docker-compose and it is still compatible.

Josca:
@beav yes, because podman-compose (until now) can only subset of features of docker-compose. But anyway I changed guides in README to podman. Usage of docker can be still supported but let's say "unofficially" :smile:

PR #94, opened 2020-01-31
Enable to call vmaas_sync - sync manually

Josca:
- also added syncing duration Prometheus metric

### DIFF
--- a/vmaas_sync/debug_api.go
+++ b/vmaas_sync/debug_api.go
@@ -1,0 +1,28 @@
+package vmaas_sync //nolint:golint,stylecheck
+
+import (
+	"app/base/utils"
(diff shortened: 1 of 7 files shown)


vmaas_sync/debug_api.go
```diff
@@ -0,0 +1,28 @@
...
+	"app/base/utils"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+func runDebugAPI() {
+	app := gin.New()
+	app.GET("/sync", func(c *gin.Context) {
+		utils.Log().Info("manual syncing called...")
+		err := syncAdvisories()
```
semtexzv: This will probably time out.

PR #97, opened 2020-02-03
System Culling

semtexzv:
- Add system culling timestamps, load them from inventory
- Add periodic jobs to vmaas_sync to delete culled systems and mark stale systems

### DIFF
--- a/listener/upload.go
+++ b/listener/upload.go
@@ -73,14 +73,16 @@ func getOrCreateAccount(account string) (int, error) {
 	return rhAccount.ID, err
 }
 
-func optParseTimestap(t *string) *time.Time {
-	if t != nil && len(*t) > 0 {
-		v, err := time.Parse(base.Rfc3339NoTz, *t)
-		if err == nil {
-			return &v
-		}
+func optParseTimestamp(t *string) *time.Time {
+	if t == nil || len(*t) > 0 {
+		return nil
 	}
-	return nil
+	v, err := time.Parse(base.Rfc3339NoTz, *t)
+	if err != nil {
+		utils.Log("err", err.Error()).Error("Opt timestamp parse")
+		return nil
+	}
+	return &v
 }
 
 // Stores or updates base system profile, returing internal system id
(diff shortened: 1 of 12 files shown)


listener/upload.go
```diff
@@ -121,46 +136,52 @@ func processUpload(hostID string, account string, identity string) error {
...
+	profileResults, _, err := inventoryClient.HostsApi.ApiHostGetHostSystemProfileById(ctx, []string{hostID}, nil)
 	if err != nil {
 		return errors.Wrap(err, "could not inventory system profile")
 	}
 
 	utils.Log().Debug("System profile download complete")
 
-	if inventoryData.Count == 0 {
-		return errors.Wrap(err, "no system details returned, host is probably deleted")
+	if profileResults.Count == 0 || hostResults.Count == 0 {
```
Josca: It's easier to debug code without these logical combinations. It's better to throw separate errors for each case. Moreover you can check `hostResults.Count` immediately after this var is obtained and not to call inventory API next time if it's zero.

listener/upload.go
```diff
@@ -121,46 +136,52 @@ func processUpload(hostID string, account string, identity string) error {
...
+	modules := make([]vmaas.UpdatesRequestModulesList, len(systemProfile.SystemProfile.DnfModules))
+	for i, m := range systemProfile.SystemProfile.DnfModules {
 		modules[i] = vmaas.UpdatesRequestModulesList{
 			ModuleName:   m.Name,
 			ModuleStream: m.Stream,
 		}
 	}
 	repos := []string{}
-	for _, r := range host.SystemProfile.YumRepos {
+	for _, r := range systemProfile.SystemProfile.YumRepos {
```
Josca: could we use `make` to preallocation here?

vmaas_sync/system_culling.go
```diff
@@ -0,0 +1,17 @@
+package vmaas_sync //nolint:golint,stylecheck
+
+import (
+	"app/base/database"
+	"time"
+)
+
+func RunSystemCulling() {
+	ticker := time.NewTicker(time.Hour)
```
Josca: This should be rather implemented as a job. We shouldn't mix different tasks in one microservice.

platform/inventory.go
```diff
@@ -65,8 +67,32 @@ func systemProfileHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, makeSystemProfile(c.Param("host_id")))
 }
 
+func systemHandler(c *gin.Context) {
```
Josca: Couldn't we replace this mock webserver with e.g. httpd serving static content?

PR #98, opened 2020-02-03
Added re-evaluation using new instance of evaluator

Josca:
- split evaluator into two components:
  - `evaluator_upload` - will evaluate system after new archive upload. Component `listener` will send kafka message for that to `patchman.evaluator.recalc` kafka topic.
  - `evaluator_recalc` - will evaluate systems after vmaas update event. Component `vmaas_sync` will send kafka messages for that (for all systems) to `patchman.evaluator.recalc` kafka topic.
- speed up `docker-compose up` development setup using var `GORUN` and local folder mounting
- updated openshift configs
- added script to `docker-compose.yml` to feed db with testing data 
- updated `README.md` with description how to test whole app


beav:
Can you give more detail on what this change does? I'm not sure what the re-evaluator's function is

Josca:
@beav ok, I added description and improved docker-compose setup to easier test it

scripts/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+COMPONENT=$1
```
semtexzv: `set -u` could be used to fail when no arguments are provided

PR #109, opened 2020-02-07
Fix listener adding disabled repos to vmaas_json

semtexzv:


listener/upload.go
```diff
@@ -176,10 +176,12 @@ func processUpload(hostID string, account string, identity string) error {
 			ModuleStream: m.Stream,
 		}
 	}
-	repos := make([]string, len(systemProfile.SystemProfile.YumRepos))
+	repos := make([]string, 0, len(systemProfile.SystemProfile.YumRepos))
```
Josca: this returns slice of size 0, so you must append, not to assign to index

PR #110, opened 2020-02-10
Send recalculate messages in batch

semtexzv:


vmaas_sync/send_messages.go
```diff
@@ -19,24 +19,25 @@ func sendReevaluationMessages() error {
 
 	ctx := context.Background()
 
-	for _, inventoryID := range inventoryIDs {
-		sendOneMessage(ctx, inventoryID)
-	}
+	sendMessages(ctx, inventoryIDs...)
```
Josca: It would be better to control batch size for sending messages. Now it will send e.g. millions of messages in one batch.

vmaas_sync/send_messages.go
```diff
@@ -19,24 +21,27 @@ func sendReevaluationMessages() error {
 
 	ctx := context.Background()
 
-	for _, inventoryID := range inventoryIDs {
-		sendOneMessage(ctx, inventoryID)
+	for i := 0; i < len(inventoryIDs)/BatchSize; i++ {
+		sendMessages(ctx, inventoryIDs[i*BatchSize:(i+1)*BatchSize]...)
```
Josca: I think that end index can be out of range and it will fail.

Josca: Anyway, what if inventoryIDs len will be less than BatchSize?

Josca:
@semtexzv there are some conflicts

PR #122, opened 2020-02-11
Added logging for empty vmaas_json to listener

Josca:


listener/upload.go
```diff
@@ -88,7 +88,11 @@ func optParseTimestamp(t *string) *time.Time {
 // Stores or updates base system profile, returing internal system id
 func updateSystemPlatform(inventoryID string, accountID int,
 	invData *inventory.HostOut, updatesReq *vmaas.UpdatesV3Request) (*models.SystemPlatform, error) {
-	updatesReqJSON, err := json.Marshal(&updatesReq)
+	updatesReqJSON, err := json.Marshal(updatesReq)
+	if string(updatesReqJSON) == `{"package_list":null}` { // TODO remove - tmp
```
semtexzv: Just check the updatesReq for invalid data, string comparison is not reliable here

Josca: I have no idea how can we get `{"package_list":null}` in database. That's why I am checking that.

PR #125, opened 2020-02-12
Added no rows affected error to listener when deleting

Josca:


listener/delete.go
```diff
@@ -31,6 +31,12 @@ func deleteHandler(event mqueue.PlatformEvent) {
 		return
 	}
 
+	if query.RowsAffected == 0 {
+		utils.Log("inventoryID", event.ID).Error("No rows modified on delete event")
```
semtexzv: I would not store this as an error. This means that the systems were not found in the database, which is a normal state. If there was an error with deletion, it was handled few lines before.

Josca: Ok, anyway it's good to distinguish if there was something deleted or not. I changed it to warn instead.

PR #127, opened 2020-02-14
Basic RBAC lite integration

semtexzv:
Blocked by:
https://github.com/RedHatInsights/insights-rbac/pull/201
https://github.com/RedHatInsights/insights-rbac/pull/206


beav:
thanks for putting in the openapi spec issues!

beav:
I haven't tested this yet but overall the logic looks good IMO

PR #131, opened 2020-02-17
Fix optTimestampParse

semtexzv:
Contained an error, which resulted in always returning null, and because of thad not saving culling info into the database.


Josca:
@semtexzv I've fixed it in #126

PR #141, opened 2020-02-20
Handle panics + log errors explicitly

semtexzv:
This should allow us to see fatal errors in external logging such as cloudwatch.

This is only a temporary solution, since panics happening in other goroutines will still just tear down the whole process.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"app/base"
 	"app/base/core"
+	"app/base/utils"
 	"app/database_admin"
 	"app/evaluator"
 	"app/listener"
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -2,17 +2,26 @@ package main
...
 	"app/evaluator"
 	"app/listener"
 	"app/manager"
 	"app/vmaas_sync"
 	"log"
 	"os"
+	"runtime/debug"
 )
 
+func handlePanics() {
```
Josca: Ok, but could we store this method somewhere in utils package? Can we write test for it?

PR #145, opened 2020-02-24
Move to host-egress queue

semtexzv:
Allows us to not call inventory at all, we can get all the info we need from kafka messages.
- Changes `mqueue.go` to only deal with kafka messages
- `events.go` now deals with standard event format for delete and eval messages.
- `upload.go` Parses messages into custom event format
- Platform mock code now uses `map[string]interface{}` instead of string interpolation

Tied with https://github.com/RedHatInsights/e2e-deploy/pull/1124

### DIFF
--- a/base/mqueue/event.go
+++ b/base/mqueue/event.go
@@ -9,8 +9,7 @@ import (
 )
 
 type PlatformEvent struct {
-	ID string `json:"id"`
-
+	ID          string  `json:"id"`
 	Type        *string `json:"type"`
 	Timestamp   *string `json:"timestamp"`
 	Account     *string `json:"account"`
@@ -18,10 +17,10 @@ type PlatformEvent struct {
 	URL         *string `json:"url"`
 }
 
-type EventHandler func(event PlatformEvent)
+type EventHandler func(message PlatformEvent)
 
 // Performs parsing of kafka message, and then dispatches this message into provided functions
-func makeKafkaHandler(eventHandler EventHandler) KafkaHandler {
+func MakeMessageHandler(eventHandler EventHandler) MessageHandler {
 	return func(m kafka.Message) {
 		var event PlatformEvent
 		err := json.Unmarshal(m.Value, &event)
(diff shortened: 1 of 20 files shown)


Josca:
@semtexzv It seems that commit message is out of date.

conf/test.env
```diff
@@ -15,7 +15,7 @@ CONSUMER_COUNT=8
 
 KAFKA_ADDRESS=platform:9092
 KAFKA_GROUP=patchman
-UPLOAD_TOPIC=platform.upload.available
+UPLOAD_TOPIC=platform.inventory.host-egress
```
Josca: I we don't depend on inventory API, we can remove INVENTORY_ADDRESS from confs listener and evaluator_common. We can also remove inventory.go module from our platform mock.

Josca: We can also remove exposed 9001 port in docker-compose.

base/mqueue/event.go
```diff
@@ -33,12 +32,12 @@ func makeKafkaHandler(eventHandler EventHandler) KafkaHandler {
...
 
-func (t *readerImpl) HandleEvents(handler EventHandler) {
-	t.HandleMessages(makeKafkaHandler(handler))
+func HandleEvents(r Reader, handler EventHandler) {
+	r.HandleMessages(MakeMessageHandler(handler))
 }
 
-// nolint: scopelint
-func (t *writerImpl) WriteEvents(ctx context.Context, events ...PlatformEvent) error {
+// nolint: scopelint, golint
```
Josca: We can satisfy golint just setting ctx as a first method argument.

listener/upload.go
```diff
@@ -10,71 +10,83 @@ import (
...
 	"github.com/pkg/errors"
+	"github.com/segmentio/kafka-go"
 	"time"
 )
 
-func uploadHandler(event mqueue.PlatformEvent) {
-	tStart := time.Now()
-	defer utils.ObserveSecondsSince(tStart, messageHandlingDuration.WithLabelValues(EventUpload))
+type Host struct {
+	ID                    string                    `json:"id,omitempty"`
```
Josca: As we need only ID, SystemProfile and culling related fields, we can ignore the others, can't we?

listener/upload.go
```diff
@@ -10,71 +10,83 @@ import (
...
+}
 
-	if event.B64Identity == nil {
-		utils.Log("inventoryID", event.ID).Error("Identity not provided")
-		messagesReceivedCnt.WithLabelValues(EventUpload, ReceivedErrorIdentity).Inc()
-		return
-	}
+type HostEgressEvent struct {
+	Type             string                 `json:"type"`
+	PlatformMetadata map[string]interface{} `json:"platform_metadata"`
```
Josca: Do we need this field?

conf/test.env
semtexzv: We can't we use it for VMaaS & RBAC mocking

listener/upload.go
semtexzv: Yes, but I'd rather follow the spec here, providing we might need some of these later

listener/upload.go
semtexzv: Not ATM, but it should contain identity and so on, I'd rather leave it here, for logging purposes at least.

listener/upload.go
Josca: I don't think so. It's better to implement only what you need. It's no problem add it later once you really need it (probably never :smile: ).

PR #159, opened 2020-03-03
Added repo-based re-evaluation - vmaas_sync

Josca:
- Implemented repo-based recalculation in vmaas_sync.
- Added `ENABLE_REPO_BASED_RE_EVALUATION` env variable.
  - It's needed to be in vmaas_sync OpenShift config before merging.

### DIFF
--- a/conf/test.env
+++ b/conf/test.env
@@ -20,3 +20,5 @@ EVENTS_TOPIC=platform.inventory.events
 EVAL_TOPIC=patchman.evaluator.upload
 EVAL_LABEL=upload
 PORT=8082
+
+ENABLE_REPO_BASED_RE_EVALUATION=true
--- a/vmaas_sync/repo_based.go
+++ b/vmaas_sync/repo_based.go
@@ -2,6 +2,9 @@ package vmaas_sync //nolint:golint,stylecheck
 import (
 	"app/base/database"
 	"app/base/models"
+	"context"
+	"github.com/RedHatInsights/patchman-clients/vmaas"
+	"github.com/antihax/optional"
 	"time"
 )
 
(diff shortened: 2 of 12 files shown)


vmaas_sync/repo_based.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+func updateRepoBaseEvalTimestampStr(value string) error {
+	err := database.Db.Exec("INSERT INTO timestamp_kv (name, value) values (?, ?)"+
+		"ON CONFLICT (name) DO UPDATE SET value = ?", LastEvalRepoBased, value, value).Error
+	return err
+}
+
+func getRepoBasedInventoryIDs(repos *[]string) (*[]string, error) {
```
semtexzv: Pointer to slice

conf/test.env
```diff
@@ -20,3 +20,5 @@ EVENTS_TOPIC=platform.inventory.events
 EVAL_TOPIC=patchman.evaluator.upload
 EVAL_LABEL=upload
 PORT=8082
+
+ENABLE_REPO_BASED_RE_EVALUATION=true
```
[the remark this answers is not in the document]
Josca: I think it's better to have parameters explicitly set to make configuration options visible.

vmaas_sync/repo_based.go
Josca: Yes, and also in other places. Is there any reason not to use it any more?

semtexzv: In go everything is copy-by-value.
You'd be right to assume that passing a slice to function would copy it, except that this only applies to fixed size array.

Passing a slice into a function only passes the slice header by value, and the actual backing array is not copied.

By creating a pointer to slice you're creating an 8byte pointer to 24byte header, which contains a pointer to actual data.

If I'm not mistaken, creating this pointer means that escape analysis will conclude that the slice will have to be allocated on heap instead of stack.

So, we have
stack -> heap -> heap
[ptr] -> [ptr, len, cap] -> [... data]

instead of 
stack -> heap
[ptr, len, cap] -> [ ... data]

Also ergonomics, we don't have to type all the dereferences and parentheses.


From what I understand, the only necessary place to use pointer to slice is when you want to modify the slice length/capacity in a function, and have that reflected in caller.

vmaas_sync/send_messages.go
```diff
@@ -12,15 +12,19 @@ import (
...
-	err := database.Db.Model(&models.SystemPlatform{}).
-		Pluck("inventory_id", &inventoryIDs).Error
+	var getInventorIDsFun func() ([]string, error)
+
+	if enabledRepoBasedReeval {
+		getInventorIDsFun = getCurrentRepoBasedInventoryIDs
+	} else {
+		getInventorIDsFun = getAllInventoryIDs
+	}
+	inventoryIDs, err := getInventorIDsFun()
```
semtexzv: You could just declare inventoryIDs & err before and not create func variable here.

PR #162, opened 2020-03-09
Increase evaluation performance

semtexzv:
- Adds a missing private key to `advisory_account_data` (main increase)
- Removes `id` column from `system_advisories`, since it is a join table
- Remove batching from kafka writer
- Misc removal of pointer-to-array antipattern


database/schema/create_schema.sql
```diff
@@ -620,7 +618,8 @@ CREATE TABLE IF NOT EXISTS advisory_account_data
     CONSTRAINT status_id
         FOREIGN KEY (status_id)
             REFERENCES status (id),
-    UNIQUE (advisory_id, rh_account_id)
+    UNIQUE (advisory_id, rh_account_id),
```
Josca: Why are we leaving UNIQUE constraint here? Couldn't we remove it like in system_advisories table because of primary key couple?

manager/controllers/advisories.go
```diff
@@ -91,7 +91,7 @@ func buildQueryAdvisories(account string) *gorm.DB {
 	return query
 }
 
-func buildAdvisoriesData(advisories []AdvisoriesDBLookup) *[]AdvisoryItem {
+func buildAdvisoriesData(advisories []AdvisoriesDBLookup) []AdvisoryItem {
```
Josca: Why are we changing that to non-pointer?

Josca: The same question for next 3 cases. Pointer usage should be more efficient because it avoids data copying.

semtexzv: Slice is implemented as a fat pointer. The only thing copied is a pointer to data, length and capacity, 24 bytes instead of 8.

database/schema/create_schema.sql
semtexzv: The unique constraint creates `advisory_id, rh-account_id` index, and the primary key creates
`rh_account_id, advisory_id` index.

PR #164, opened 2020-03-10
Added repo-system saving to listener

Josca:


listener/upload.go
```diff
@@ -148,24 +149,109 @@ func updateSystemPlatform(inventoryID string, accountID int, host *Host,
...
+}
+
+func ensureReposInDb(tx *gorm.DB, repos []string) (repoIDs []int, added int64, err error) {
+	repoObjs := make(models.RepoSlice, len(repos))
+	for i, repo := range repos {
+		repoObjs[i] = models.Repo{Name: repo}
+	}
+
+	txOnConflict := tx.Set("gorm:insert_option", "ON CONFLICT DO NOTHING")
+	err = database.BulkInsert(txOnConflict, repoObjs)
```
semtexzv: We need to investigate whether selecting the repos & doing filtering on client side is faster or not

PR #165, opened 2020-03-11
Optimize listener message handling

semtexzv:
- Wrap code in single transaction, previously it would run in 2 transactions
- Skip sending eval messages for systems, which haven't had their `system_profile` changed.
- Skip re-writing `vmaas_json` if it haven't changed

### DIFF
--- a/listener/upload.go
+++ b/listener/upload.go
@@ -145,10 +145,8 @@ func updateSystemPlatform(tx *gorm.DB, inventoryID string, accountID int, host *
 	}
 
 	hash := sha256.New()
-	_, err = hash.Write(updatesReqJSON)
-	if err != nil {
-		return nil, errors.Wrap(err, "Unable to hash updates json")
-	}
+	// Never returns an error
+	hash.Write(updatesReqJSON) // nolint: errcheck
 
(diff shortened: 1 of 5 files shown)


listener/upload.go
```diff
@@ -86,24 +88,38 @@ func uploadHandler(event HostEgressEvent) {
...
 
-	err := processUpload(context.Background(), event.Host.ID, event.Host.Account, &event.Host)
+	sys, err := processUpload(event.Host.Account, &event.Host)
 	if err != nil {
 		utils.Log("inventoryID", event.Host.ID, "err", err.Error()).Error(ErrorProcessUpload)
 		messagesReceivedCnt.WithLabelValues(EventUpload, ReceivedErrorProcessing).Inc()
 		return
 	}
 
+	if sys.UnchangedSince != nil && sys.LastEvaluation != nil {
```
Josca: Could we add test for that? It's not covered in tests and so we are slowly decreasing tests coverage:
https://codecov.io/gh/RedHatInsights/patchman-engine/pull/165/src/listener/upload.go?before=listener/upload.go

PR #185, opened 2020-03-24
Fix cache calculation

semtexzv:
There were several issues which caused invalid caches.
- Not locking `advisory_account_data` in between 2 updates when evaluation could result in an inconsistencies when evaluating 2 package profiles with same advisories from same account
- Not locking `system_platform` rows when performing system culling tasks
- Not locking `system_platform` when inserting from listener

### DIFF
--- a/base/database/testing.go
+++ b/base/database/testing.go
@@ -2,10 +2,91 @@ package database
 
 import (
 	"app/base/models"
+	"app/base/utils"
 	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 )
+
+func DebugWithCachesCheck(part string, fun func()) {
+	fun()
+	validAfter, err := CheckCachesValid()
+	if err != nil {
+		utils.Log("error", err).Error("Could not check validity of caches")
+	}
+
+	if !validAfter {
+		utils.Log("part", part).Panic("Cache mismatch created")
+	}
+}
+
+type key struct {
+	AccountID  int
+	AdvisoryID int
+}
+
+type advisoryCount struct {
+	RhAccountID int
+	AdvisoryID  int
+	Count       int
+}
+
+func CheckCachesValid() (bool, error) {
+	valid := true
+	var aad []models.AdvisoryAccountData
+
+	tx := Db.Begin()
+	err := tx.Set("gorm:query_option", "FOR SHARE OF advisory_account_data").
+		Order("rh_account_id, advisory_id").Find(&aad).Error
+	if err != nil {
+		return false, err
+	}
+	var counts []advisoryCount
+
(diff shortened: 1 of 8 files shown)


base/database/testing.go
```diff
@@ -2,11 +2,91 @@ package database
 
 import (
 	"app/base/models"
+	"app/base/utils"
 	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 )
 
+func DebugWithCachesCheck(part string, fun func()) {
```
Josca: This is not used anywhere. It would be useful to allow enabling this checking after evaluation using env var (ENABLE_CACHES_CHECK, default false). But we could reduce checking on single account only.

base/database/testing.go
```diff
@@ -2,11 +2,91 @@ package database
...
+	if err != nil {
+		utils.Log("error", err).Error("Could not check validity of caches")
+	}
+
+	if !validAfter {
+		utils.Log("part", part).Panic("Cache mismatch created")
+	}
+}
+
+// nolint: funlen
```
Josca: This lint is unneccessarry. It's better to split method into multiple submethods. That's why we use the lint :wink:.

base/database/testing.go
```diff
@@ -2,11 +2,91 @@ package database
...
+	for key, calcCount := range calculated {
+		cachedCount := calculated[key]
+
+		if cachedCount != calcCount {
+			utils.Log("advisory_id", key.AdvisoryID, "account_id", key.AccountID,
+				"cached", cachedCount, "calculated", calcCount).Error("Cached counts mismatch")
+			valid = false
+		}
+	}
+	tx.Commit()
```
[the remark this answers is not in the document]
semtexzv: If we don't commit the transaction, it keeps an open slot on the server blocking other transactions

evaluator/evaluate.go
```diff
@@ -347,31 +346,43 @@ func ensureAdvisoryAccountDataInDb(tx *gorm.DB, rhAccountID int, advisoryIDs []i
 	return err
 }
 
-func updateSystemAdvisories(tx *gorm.DB, systemID, rhAccountID int, patched, unpatched []int) error {
+func updateSystemAdvisories(tx *gorm.DB, system *models.SystemPlatform, patched, unpatched []int) error {
```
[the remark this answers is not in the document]
semtexzv: Intended for future change. I think we can calculate advisory_type counts on the client.

base/database/testing.go
semtexzv: I added the method for tracking down bugs and kept it for test purposes. I don't want to run it in production environments, since it does not provide enough information.

Invalid caches appeared due to concurrency issues, and method was pretty much useless for tracking down where the issue was.

PR #193, opened 2020-03-27
Implement retrying for message handling

semtexzv:


base/mqueue/mqueue.go
```diff
@@ -65,7 +66,22 @@ func WriterFromEnv(topic string) Writer {
...
+func MakeRetryingHandler(handler MessageHandler) MessageHandler {
+	return func(message kafka.Message) error {
+		var err error
+		b, cancel := policy.Start(context.Background())
+		defer cancel()
+		for backoff.Continue(b) {
+			if err = handler(message); err == nil {
+				return nil
+			}
+			utils.Log("err", err.Error()).Error("Try failed")
```
Josca: Maybe I would include retry number.

base/mqueue/mqueue.go
```diff
@@ -65,7 +66,22 @@ func WriterFromEnv(topic string) Writer {
 	return writer
 }
 
-type MessageHandler func(message kafka.Message)
+type MessageHandler func(message kafka.Message) error
+
+func MakeRetryingHandler(handler MessageHandler) MessageHandler {
+	return func(message kafka.Message) error {
+		var err error
+		b, cancel := policy.Start(context.Background())
```
Josca: I didn't know what b is and needed to find it in code. I would rather to name it backoff.

PR #207, opened 2020-04-01
added debug "caches-check" endpoint

Josca:


scripts/caches-check.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# script to manually call vmaas_sync
+
+curl http://localhost:9999/caches-check
```
semtexzv, line 5: I'd name it check-caches, since it is an action performed, not a resource.

Josca: fixed

PR #210, opened 2020-04-03
Calculate system caches in code

semtexzv:
- Does not use database function for calculating system caches
- Updates last_evaluation when updating caches, removing one write per evaluation

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -20,6 +20,8 @@ const (
 	unknown = "unknown"
 )
 
+type SystemAdvisoryMap map[string]models.SystemAdvisories
+
 var (
 	consumerCount int
 	vmaasClient   *vmaas.APIClient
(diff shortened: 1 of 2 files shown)


Josca:
@semtexzv tests failed on wrong go format.

go.sum
```diff
@@ -91,6 +91,7 @@ github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7
 github.com/erikstmartin/go-testdb v0.0.0-20160219214506-8d10e4a1bae5/go.mod h1:a2zkGnVExMxdzMo3M0Hi/3sEU+cWnZpSni0O6/Yb/P0=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/fsouza/fake-gcs-server v1.17.0/go.mod h1:D1rTE4YCyHFNa99oyJJ5HyclvN/0uQR+pM/VdlL83bw=
+github.com/getkin/kin-openapi v0.3.0 h1:xsQ4mA20YJDMgIHdHqMKZ66QUe6/hi+x6yLsTTz8xyQ=
```
Josca, line 94: why do we need to update this file?

evaluator/evaluate.go
```diff
@@ -100,48 +102,54 @@ func commitWithObserve(tx *gorm.DB) error {
...
-func updateCaches(tx *gorm.DB, system *models.SystemPlatform) error {
-	tStart := time.Now()
-	defer utils.ObserveSecondsSince(tStart, evaluationPartDuration.WithLabelValues("caches-update"))
-
-	err := tx.Exec("SELECT * FROM refresh_system_caches(?,?)", system.ID, system.RhAccountID).Error
+	for _, sa := range old {
+		// TODO: Add dedicated counter to unknown advisories
+		if sa.WhenPatched == nil && sa.Advisory.AdvisoryTypeID > 0 {
+			counts[sa.Advisory.AdvisoryTypeID]++
+		}
```
Josca, line 142: Is SQL `refresh_system_caches(?,?)` method used anywhere now? If not, could we add note "TODO remove" to it or so?

evaluator/evaluate.go
```diff
@@ -100,48 +102,54 @@ func commitWithObserve(tx *gorm.DB) error {
 }
 
 func evaluateAndStore(tx *gorm.DB, system *models.SystemPlatform, vmaasData vmaas.UpdatesV2Response) error {
-	patched, unpatched, err := processSystemAdvisories(tx, system, vmaasData, system.InventoryID)
+	oldSystemAdvisories, patched, unpatched, err := processSystemAdvisories(tx, system, vmaasData, system.InventoryID)
```
Josca, line 105: I see three names for this value: oldSystemAdvisories (here), advisories (method header), stored (when returning in method). Could we unify it, please?

evaluator/evaluate.go
```diff
@@ -100,48 +102,54 @@ func commitWithObserve(tx *gorm.DB) error {
...
-	}
 	return nil
 }
 
-func updateSystemLastEvaluation(tx *gorm.DB, system *models.SystemPlatform) error {
+func updateSystemPlatform(tx *gorm.DB, system *models.SystemPlatform, old, new SystemAdvisoryMap) error {
 	tStart := time.Now()
-	defer utils.ObserveSecondsSince(tStart, evaluationPartDuration.WithLabelValues("last-evaluation-update"))
+	defer utils.ObserveSecondsSince(tStart, evaluationPartDuration.WithLabelValues("system-update"))
+	if old == nil || new == nil {
```
Josca, line 129: I did not find the case where old or new could be nil. Is there any?

semtexzv: Happened during some test cases.

go.sum
semtexzv: No ida, put in by `go mod tidy`

evaluator/evaluate.go
semtexzv: I'd like to keep it for future purposes, refreshing system caches from psql can be useful.

PR #212, opened 2020-04-06
Add tests to system culling, update opt_out procedure

semtexzv:


database_admin/migrations/018_caches_fix.up.sql
```diff
@@ -0,0 +1,114 @@
+CREATE OR REPLACE FUNCTION mark_stale_systems()
```
Josca, line 1: Shouldn't we add previous method version into down.sql script?

database_admin/schema/create_schema.sql
```diff
@@ -159,6 +156,7 @@ BEGIN
     FROM system_advisories sa
     WHERE sa.system_id = NEW.id
       AND sa.when_patched IS NULL
+      AND change > 0
```
Josca, line 159: Just to be sure, isn't it be `change != 0`?

database_admin/migrations/018_caches_fix.up.sql
semtexzv: Previous version had a significant mistake, I don't see a reason to have it around.

database_admin/schema/create_schema.sql
semtexzv: No, this is a logical error. It should never happen, but in the old code we performed insertion only when incrementing the counters.

PR #215, opened 2020-04-07
Ensured gracefull containers termination

Josca:


base/database/setup.go
```diff
@@ -20,6 +20,19 @@ func Configure() {
 	check(Db)
 }
 
+func Close() {
```
semtexzv, line 23: Can you add a test for this ? Maybe running `pg_sleep` in a transaction and Closing the connection from a goroutine.

Josca: ok, will try it

base/database/setup.go
```diff
@@ -75,18 +73,17 @@ func loadEnvPostgreSQLConfig() *PostgreSQLConfig {
 		Passwd:   utils.Getenv("DB_PASSWD", "FILL"),
 		SSLMode:  utils.Getenv("DB_SSLMODE", "FILL"),
 
-		Timeout:               "60s",
-		ReadTimeout:           "60s",
-		WriteTimeout:          "60s",
-		MaxConnections:        250,
-		MaxIdleConnections:    50,
-		MaxConnectionLifetime: 60,
+		StatementTimeoutMs:     utils.GetIntEnvOrDefault("DB_STATEMENT_TIMEOUT_MS", 10000),
```
semtexzv, line 76: This won't work. The system culling statements definitely take more than 10 seconds

base/database/testing.go
```diff
@@ -36,7 +37,7 @@ func CheckCachesValid() (bool, error) {
 	valid := true
 	var aad []models.AdvisoryAccountData
 
-	tx := Db.Begin()
+	tx := Db.BeginTx(base.Context, nil)
```
semtexzv, line 40: We're also using context in kafka reader & writer implementation & API calls. It'd be nice to use drerived contexts there, which would allow us to stop all processing when receiving a signal.

base/database/setup.go
Josca: We can turn it of setting it to 0 (https://www.postgresql.org/docs/10/runtime-config-client.html) for some components, e.g. vmaas_sync or increase the value. But it's useful to avoid cumulating dead transactions in case that container is killed.

Josca: Or we can turn it off by default (setting to `0`) and set timeouts only for some components (listener, evaluator).

base/database/testing.go
Josca: Ok, good idea.

PR #218, opened 2020-04-08
Manual updating of stale flag in listener

semtexzv:
The `on_system_timestamp_update` did not call the trigger to update cached counts after updating the `stale` flag.
This changes so that listener updates this flag manually.

### DIFF
--- a/listener/upload.go
+++ b/listener/upload.go
@@ -157,6 +157,7 @@ func updateSystemPlatform(tx *gorm.DB, inventoryID string, accountID int, host *
 	var colsToUpdate = []string{
 		"display_name",
 		"last_upload",
+		"stale",
 		"stale_timestamp",
 		"stale_warning_timestamp",
 		"culled_timestamp",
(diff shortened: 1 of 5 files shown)


listener/upload.go
```diff
@@ -178,6 +181,7 @@ func updateSystemPlatform(tx *gorm.DB, inventoryID string, accountID int, host *
 		StaleTimestamp:        optParseTimestamp(host.StaleTimestamp),
 		StaleWarningTimestamp: optParseTimestamp(host.StaleWarningTimestamp),
 		CulledTimestamp:       optParseTimestamp(host.CulledTimestamp),
+		Stale:                 staleWarning != nil && staleWarning.Before(time.Now()),
```
Josca, line 184: Is it good place to update Stale timestamp? I thought stale systems are those we don't get uploads for.

semtexzv: Probably, only other place where we save systems is at the end of evaluation.

semtexzv: Stale systems are the ones, which haven't received upload in a while. That does not mean they can't get uploaded.

PR #220, opened 2020-04-14
Systems stat

Josca:
Updated systems counts metrics with labels: `opt_out`, `last_upload`.


vmaas_sync/metrics.go
```diff
@@ -96,26 +99,58 @@ func update() {
...
 func updateSystemMetrics() {
-	optOuted, notOptOuted, err := getSystemCounts()
+	counts, err := getSystemCounts(time.Now())
 	if err != nil {
 		utils.Log("err", err.Error()).Error("unable to update system metrics")
+		return
+	}
+
+	for key, count := range counts {
+		labels := strings.Split(key, ":")
```
semtexzv, line 109: Why not use a simple struct here instead of concatenating & splitting strings ?

vmaas_sync/metrics.go
```diff
@@ -96,26 +99,58 @@ func update() {
...
-	if err != nil {
-		return 0, 0, errors.Wrap(err, "unable to get not opt_outed systems")
+// Update input systems query with "opt_out = X" constraint.
+func updateSystemsQueryOptOut(systemsQuery *gorm.DB, optOut bool) *gorm.DB {
+	return systemsQuery.Where("opt_out = ?", optOut)
+}
+
+// Update input systems query with "last_upload > T" constraint.
+// Constraint is not added if "lastNDays" argument is negative.
+func updateSystemsQueryLastUpload(systemsQuery *gorm.DB, refTime time.Time, lastNDays int) *gorm.DB {
```
semtexzv, line 149: Weird naming, maybe use  `withLastUploadFilter` ?

Josca: I used prefix notation here updateSystemsQuery<column-name>. Similarly in case of opt out. I didn't use "with" to make it shorter.

vmaas_sync/metrics.go
Josca: yes, good idea, thanks, done

PR #224, opened 2020-04-15
Better error handling for system culling

semtexzv:


vmaas_sync/system_culling.go
```diff
@@ -4,23 +4,41 @@ import (
...
 )
 
+// Need to run code within a function, because defer can't be used in loops
+func withTx(do func(db *gorm.DB) error) error {
+	tx := database.Db.BeginTx(base.Context, nil)
+	defer tx.RollbackUnlessCommitted()
+	if err := do(tx); err != nil {
+		return err
+	}
+	return errors.Wrap(tx.Commit().Error, "Commit")
```
Josca, line 19: that's a little confusing, we rely on errors.Wrap returns Nil on Nil input argument.

PR #226, opened 2020-04-16
Added Kafka readers closing using WaitGroup

Josca:
- also added testing upload loop

### DIFF
--- a/base/mqueue/mqueue.go
+++ b/base/mqueue/mqueue.go
@@ -8,8 +8,11 @@ import (
 	"github.com/lestrrat-go/backoff"
 	"github.com/segmentio/kafka-go"
 	"io"
+	"sync"
 	"time"
 )
+
+const errContextCanceled = "context canceled"
 
 // By wrapping raw value we can add new methods & ensure methods of wrapped type are callable
 type Reader interface {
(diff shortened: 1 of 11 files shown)


base/mqueue/mqueue.go
```diff
@@ -91,6 +92,9 @@ func (t *readerImpl) HandleMessages(handler MessageHandler) {
 	for {
 		m, err := t.FetchMessage(base.Context)
 		if err != nil {
+			if err.Error() == "context canceled" {
```
semtexzv, line 95: Please extract the string to a constant

Josca: done

evaluator/evaluate.go
```diff
@@ -452,11 +453,14 @@ func run(readerBuilder mqueue.CreateReader) {
 	// We create multiple consumers, and hope that the partition rebalancing
 	// algorithm assigns each consumer a single partition
 	for i := 0; i < consumerCount; i++ {
-		go mqueue.RunReader(evalTopic, readerBuilder, handler)
+		wg.Add(1)
```
semtexzv, line 456: Why increment here ? The RunReader should manage its contribution to the waitgroup.

Josca: If we put it there (inside go routine function) theoretically wg.Wait() calling in main thread could be reached earlier than first wg.Add(1) is called.

PR #230, opened 2020-04-17
Safe system deletes

semtexzv:


listener/upload.go
```diff
@@ -333,14 +341,27 @@ func processUpload(account string, host *Host) (*models.SystemPlatform, error) {
...
+
+	var deleted models.DeletedSystem
+	if err := tx.Find(&deleted, "inventory_id = ?", host.ID).Error; err != nil &&
+		!gorm.IsRecordNotFoundError(err) {
+		return nil, errors.Wrap(err, "Checking deleted systems")
+	}
+
+	// If the system was deleted in last hour, don't register this upload
+	if deleted.InventoryID != "" && deleted.WhenDeleted.After(time.Now().Add(-DeletionThreshold)) {
+		utils.Log("inventoryID", host.ID).Info("Received recently deleted system")
```
Josca, line 354: any test for that case?

listener/upload.go
```diff
@@ -25,6 +25,8 @@ const (
 	ErrorProcessUpload     = "unable to process upload"
 	UploadSuccessNoEval    = "upload event handled successfully, no eval required"
 	UploadSuccess          = "upload event handled successfully"
+
+	DeletionThreshold = 4 * time.Hour
```
Josca, line 29: would be better to create env var for that with 4h default value

database_admin/migrations/020_delete_cache.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    when_deleted TIMESTAMP WITH TIME ZONE NOT NULL,
+    UNIQUE (inventory_id)
+) TABLESPACE pg_default;
+
+CREATE INDEX ON deleted_system (when_deleted);
+
+GRANT SELECT, INSERT, UPDATE, DELETE ON deleted_system TO listener;
+GRANT SELECT ON TABLE deleted_system TO evaluator;
+GRANT SELECT ON TABLE deleted_system TO manager;
+GRANT SELECT ON TABLE deleted_system TO vmaas_sync;
```
Josca, line 14: Maybe we should add delete permission to vmaas_sync and set up job to delete old deleted systems into vmaas_sync, right?

listener/upload.go
semtexzv: Added test in delete_test.go

listener/upload.go
semtexzv: Thanks, fixed

database_admin/migrations/020_delete_cache.up.sql
semtexzv: I've gone with the way that vulnerability does it and perform cleanup of the deleted_system table in listener.

PR #237, opened 2020-04-23
Added support to run component locally

Josca:


docker-compose.yml
```diff
@@ -42,19 +42,38 @@ services:
...
 
+  zookeeper:
+    image: confluentinc/cp-zookeeper:5.4.0
+    container_name: zookeeper
+    env_file:
+      - ./conf/zookeeper.env
+    ports:
+      - 2181:2181
+
+  kafka:
```
semtexzv, line 53: Couldn't we run an image with contains both zookeper and kafka ?

platform/platform.go
```diff
@@ -14,7 +14,7 @@ import (
 var websockets []chan string
 var runUploadLoop = false
 
-func AddWebsocket() chan string {
+func addWebsocket() chan string {
```
semtexzv, line 17: Why are all these public ? We are calling them only in the `platform` package.

docker-compose.yml
Josca: Yes, but we would have to build it ourself. This is standard way according to kafka docs.

platform/platform.go
Josca: This is question for you, you wrote this code. I made it private right in this PR.

PR #239, opened 2020-04-24
Added kafka error Prometheus counter

Josca:
- Added prometheus counters counting Kafka connection read/write errors.
- Fixed TestRunServer test.
- Fixed and unified Kafka connection errors logging.

### DIFF
--- a/base/mqueue/metrics.go
+++ b/base/mqueue/metrics.go
@@ -1,0 +1,22 @@
+package mqueue
+
+type Counter interface {
+	Inc()
+}
+
+var (
+	kafkaErrorReadCnt  Counter = &emptyCnt{}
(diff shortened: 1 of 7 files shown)


base/mqueue/metrics.go
```diff
@@ -0,0 +1,22 @@
+package mqueue
+
+type Counter interface {
```
semtexzv, line 3: Why do we need to make this abstraction ?

Josca: This avoids adding new dependencies into modules if they are not directly needed here. We can add prometheus library dependency in the same place where metrics module usage is called. This way mqueue does not depend on Prometheus library.

PR #240, opened 2020-04-27
Run postgres with logging_collector disabled for dev

semtexzv:


docker-compose.test.yml
```diff
@@ -5,6 +5,7 @@ services:
     build:
       context: .
       dockerfile: dev/database/Dockerfile.centos
+    command: 'run-postgresql -c logging_collector=off'
```
Josca, line 8: why do we add this?

semtexzv: When running locally, we don't get any logs, since they are redirected to file by logging collector. This change will leave them on stderr instead of redirecting to file.

Josca: aha, ok

PR #249, opened 2020-05-05
Upgraded database from postgres 10 to 12

Josca:


semtexzv:
Did you try to upgrade pre-existing database ? tests run on fresh database every time, so we don't know whether the CI/QA envs will be upgraded correctly

Josca:
@semtexzv yes, I tried that now. We will need to run ugraded db container with POSTGRESQL_UPGRADE=copy (https://hub.docker.com/r/centos/postgresql-12-centos7). I tried it in docker with exisiting volume ver10 and upgrade to version 12.

semtexzv:
Can be merged after https://github.com/RedHatInsights/e2e-deploy/pull/2287

PR #251, opened 2020-05-11
Fix listener crashing when receiving updated messages with missing systems

semtexzv:


listener/events.go
```diff
@@ -55,7 +55,10 @@ func EventsMessageHandler(m kafka.Message) error {
 
 func HandleUpdate(event HostEgressEvent) error {
 	var system models.SystemPlatform
-	if err := database.Db.Find(&system, "inventory_id = ?", event.Host.ID).Error; err != nil {
+	err := database.Db.Find(&system, "inventory_id = ?", event.Host.ID).Error
+	if err != nil && gorm.IsRecordNotFoundError(err) {
+		return nil
```
Josca, line 60: what to log it at least?

semtexzv: Thanks. I added metrics to different code paths.

listener/events.go
```diff
@@ -55,16 +55,24 @@ func EventsMessageHandler(m kafka.Message) error {
...
 func HandleUpdate(event HostEgressEvent) error {
 	var system models.SystemPlatform
-	if err := database.Db.Find(&system, "inventory_id = ?", event.Host.ID).Error; err != nil {
+	err := database.Db.Find(&system, "inventory_id = ?", event.Host.ID).Error
+	if err != nil && gorm.IsRecordNotFoundError(err) {
+		utils.Log("inventoryID", event.Host.ID).Info("System not found when handling update")
+		messagesReceivedCnt.WithLabelValues(EventUpdate, ReceivedErrorNoRows).Inc()
+		return nil
+	} else if err != nil {
+		messagesReceivedCnt.WithLabelValues(EventUpdate, ReceivedErrorOtherType).Inc()
```
Josca, line 64: I would add error logs even to these two error places. This is the only way how to see what happens in case of some troubles, right?

semtexzv: Well these errors will be propagated and handled at the level above.

semtexzv: Changed plain errors to wrapped errors.

Josca: ok

PR #256, opened 2020-05-13
Add a metric to report errors when calling other services

semtexzv:


manager/middlewares/rbac.go
```diff
@@ -5,9 +5,17 @@ import (
 	"app/base/utils"
 	"github.com/RedHatInsights/patchman-clients/rbac"
 	"github.com/gin-gonic/gin"
+	"github.com/prometheus/client_golang/prometheus"
 	"net/http"
+	"strconv"
 )
 
+var serviceErrorCnt = prometheus.NewCounterVec(prometheus.CounterOpts{
```
Josca, line 13: As it's intended also for other services it should be defined in middlewares/prometheus.go file.

manager/middlewares/rbac.go
```diff
@@ -24,10 +32,18 @@ func isAccessGranted(c *gin.Context) bool {
...
 	// nolint: bodyclose
-	access, _, err := client.AccessApi.GetPrincipalAccess(base.Context, "patch", nil)
+	access, res, err := client.AccessApi.GetPrincipalAccess(base.Context, "patch", nil)
+	if res != nil && res.Body != nil {
+		defer res.Body.Close()
+	}
 
 	if err != nil {
 		utils.Log("err", err.Error()).Error("Call to RBAC svc failed")
+		status := 500
```
Josca, line 42: I would rather use http.InternalServerError constant here.

Josca:
@semtexzv it failed.

PR #259, opened 2020-05-20
Added optional KAFKA_READER_MIN_BYTES and KAFKA_READER_MAX_BYTES vars

Josca:
There were some issues due to Kafka settings in perf-testing environment. Kafka reader config update solve this problem. I've added options to change Kafka reader config using env vars.

### DIFF
--- a/base/mqueue/mqueue.go
+++ b/base/mqueue/mqueue.go
@@ -35,16 +35,17 @@ type writerImpl struct {
 func ReaderFromEnv(topic string) Reader {
 	kafkaAddress := utils.GetenvOrFail("KAFKA_ADDRESS")
 	kafkaGroup := utils.GetenvOrFail("KAFKA_GROUP")
+	minBytes := utils.GetIntEnvOrDefault("KAFKA_READER_MIN_BYTES", 1)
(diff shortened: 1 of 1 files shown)


base/mqueue/mqueue.go
```diff
@@ -35,16 +35,17 @@ type writerImpl struct {
 func ReaderFromEnv(topic string) Reader {
 	kafkaAddress := utils.GetenvOrFail("KAFKA_ADDRESS")
 	kafkaGroup := utils.GetenvOrFail("KAFKA_GROUP")
+	minBytes := utils.GetIntEnvOrDefault("KAFKA_READER_MIN_BYTES", 1)
+	maxBytes := utils.GetIntEnvOrDefault("KAFKA_READER_MAX_BYTES", 1e6)
```
semtexzv, line 39: Seems fine, But we need to e cautious, since the MAX_BYTES variable will siginficantly affect memory usage

Josca: There is no reason to change the value in production, but it turned out in perf-testing it's better to have a possibility to change it.

PR #262, opened 2020-05-25
Implement filters for export endpoints

semtexzv:
- Adds a way to filter exported data, format is preserved between API and export endpoints


manager/controllers/advisories.go
```diff
@@ -100,6 +100,14 @@ func AdvisoriesListHandler(c *gin.Context) {
 // @Security RhIdentity
 // @Accept   json
 // @Produce  json,text/csv
+// @Param    search         query   string  false   "Find matching text"
```
Josca, line 103: Could we align these columns or the rows would be too long?

PR #264, opened 2020-05-27
Added ENABLE_VMAAS_CALL_COMPRESSION option to evaluator

Josca:


conf/evaluator_common.env
```diff
@@ -1,2 +1,3 @@
 DB_USER=evaluator
 DB_PASSWD=evaluator
+VMAAS_CALL_COMPRESSION=on
```
semtexzv, line 3: We should follow naming convention where booleans have `ENABLE` in the name

Josca: updated

PR #273, opened 2020-06-17
Send updates to remediations after evaluation

semtexzv:
Depends on https://github.com/RedHatInsights/e2e-deploy/pull/1853

### DIFF
--- a/conf/evaluator_common.env
+++ b/conf/evaluator_common.env
@@ -1,3 +1,4 @@
 DB_USER=evaluator
 DB_PASSWD=evaluator
 ENABLE_VMAAS_CALL_COMPRESSION=true
+REMEDIATIONS_UPDATE_TOPIC=platform.application.remediation-updates
--- a/conf/evaluator_common.env
+++ b/conf/evaluator_common.env
@@ -1,4 +1,4 @@
 DB_USER=evaluator
 DB_PASSWD=evaluator
 ENABLE_VMAAS_CALL_COMPRESSION=true
-REMEDIATIONS_UPDATE_TOPIC=platform.application.remediation-updates
+REMEDIATIONS_UPDATE_TOPIC=platform.remediation-updates.patch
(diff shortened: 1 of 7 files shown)


evaluator/remediations.go
```diff
@@ -0,0 +1,28 @@
...
+
+type RemediationsState struct {
+	HostID string   `json:"host_id"`
+	Issues []string `json:"issues"`
+}
+
+func publishRemediationsState(id string, response vmaas.UpdatesV2Response) error {
+	var state RemediationsState
+	state.HostID = id
+	state.Issues = make([]string, 0)
```
Josca, line 19: We call `getReportedAdvisories` twice during evaluation it would be better to reuse its result, but I see it would make our code more complex. So at least we can create "reporetedAdvisories" first and then preallocate state.Issues with the exact length.

evaluator/remediations.go
```diff
@@ -0,0 +1,28 @@
+package evaluator
+
+import (
+	"app/base"
+	"app/base/mqueue"
+	"github.com/RedHatInsights/patchman-clients/vmaas"
+)
+
+var remediationsPublisher = mqueue.WriterFromEnv("REMEDIATIONS_UPDATE_TOPIC")
```
Josca, line 9: I would consider to make this sending optional. When REMEDIATIONS_UPDATE_TOPIC is empty, do not send anything. It can be useful for debugging and when running in testing environment when there is no consumer for these messages. Actually does it work in our local setup? Don't we need to create this topic in setup.sh?

conf/evaluator_common.env
```diff
@@ -1,3 +1,4 @@
 DB_USER=evaluator
 DB_PASSWD=evaluator
 ENABLE_VMAAS_CALL_COMPRESSION=true
+REMEDIATIONS_UPDATE_TOPIC=platform.application.remediation-updates
```
Josca, line 4: end line

evaluator/remediations.go
semtexzv: Fixed

PR #276, opened 2020-06-18
Added "packages" info into the advisory detail endpoint

Josca:


docs/openapi.json
```diff
@@ -931,6 +931,9 @@
                     "modified_date": {
                         "type": "string"
                     },
+                    "packages": {
+                        "type": "string"
```
semtexzv, line 935: This is not correct, we need to either create custom wrapper for JSONB, or use the actual AvisoryPackageDat struct here.

Josca: fixed, also fixed CVEs format

PR #277, opened 2020-06-22
Implement system tagging

semtexzv:
- Adds locally stored tags
- Allows implements tag filtering on /systems endpoints
- Does not perform deduplication of tags by moving them to separate table. Current implementation has simpler evaluation.


database_admin/migrations/025_tags.up.sql
```diff
@@ -0,0 +1,35 @@
+
+/*CREATE TABLE tag
```
Josca, line 2: forgotten comments here

database_admin/migrations/025_tags.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE system_tag CASCADE;
+DROP TABLE tag CASCADE;
```
Josca, line 2: end line

database_admin/migrations/025_tags.up.sql
```diff
@@ -0,0 +1,35 @@
...
+GRANT SELECT ON tag to manager;
+
+GRANT SELECT, USAGE ON SEQUENCE public.tag_id_seq TO evaluator;
+GRANT SELECT, USAGE ON SEQUENCE public.tag_id_seq TO listener;
+GRANT SELECT, USAGE ON SEQUENCE public.tag_id_seq TO vmaas_sync;
+*/
+
+CREATE TABLE system_tags
+(
+    tag       TEXT NOT NULL CHECK ( NOT EMPTY(tag)),
```
Josca, line 24: Why not to add tags table?

listener/common_test.go
```diff
@@ -70,10 +70,26 @@ func getOrCreateTestAccount(t *testing.T) int {
...
 	ev := HostEgressEvent{
 		Type:             "created",
 		PlatformMetadata: nil,
-		Host:             Host{ID: inventoryID, Account: inventoryID},
+		Host: Host{
+			ID:      inventoryID,
+			Account: inventoryID,
+			Tags: []inventory.StructuredTag{
+				{
+					Key:       "env",
```
Josca, line 83: Isn't better to use testing values "key1, key2, ns1, ns2, val1, val2" to easy distinguish it in the tests?

listener/upload.go
```diff
@@ -354,14 +354,45 @@ func processUpload(account string, host *Host) (*models.SystemPlatform, error) {
...
+	}
 	err = tx.Commit().Error
 	if err != nil {
 		return nil, errors.Wrap(err, "Committing changes")
 	}
 	return sys, nil
 }
+
+func updateTags(tx *gorm.DB, systemID int, host *Host) error {
+	var tags []string
```
Josca, line 373: preallocation possible here

manager/controllers/package_systems.go
```diff
@@ -34,14 +34,17 @@ func PackageSystemsListHandler(c *gin.Context) {
...
 		Table("system_platform").
 		Select("system_platform.inventory_id, package_data -> ? -> 'version' ->> 0 as version", packageName).
 		Joins("inner join rh_account ra on system_platform.rh_account_id = ra.id").
 		Where("ra.name = ?", account).
 		// TODO: this seems to not be accelerated by the gin index, investigate
-		Where("jsonb_exists(system_platform.package_data, ?)", packageName).
-		Scan(&systems).Error
+		Where("jsonb_exists(system_platform.package_data, ?)", packageName)
+
+	query, _ = ApplyTagsFilter(c, query, "system_platform.id")
```
Josca, line 46: Why the second return val is ignored?

manager/controllers/utils.go
```diff
@@ -197,6 +197,25 @@ func ApplySearch(c *gin.Context, tx *gorm.DB, searchColumns ...string) *gorm.DB
 	return txWithSearch
 }
 
+// Filter systems by tags,
+// TODO: Implement implicit namespaces
+// TODO: we currently use same encoding inside the DB, as is used in query strings
+// the namespace/name=value encoding. Ensure this is correct by parsing and re-formatting the values
+func ApplyTagsFilter(c *gin.Context, tx *gorm.DB, systemIDExpr string) (*gorm.DB, bool) {
```
Josca, line 204: I don't see the second return val (bool) to be used.

semtexzv: It'll be used for switching to manual data calculation in case of `/advisories` endpoint.

manager/controllers/package_systems.go
semtexzv: Not needed here

database_admin/migrations/025_tags.up.sql
semtexzv: I've tried but I'm hesitant to do it this way because tags are pretty short, and evaluation + querying is much simpler this way. If we ever need to move to a system where we have tags in separate table, it'll be a simple migration.

PR #282, opened 2020-06-30
Database changes for consistent package data

semtexzv:
Store package data in a relational schema, for supporting showing and performign operations on per-package information like summaries & descriptions.


database_admin/migrations/025_packages.up.sql
```diff
@@ -0,0 +1,54 @@
...
+GRANT SELECT, USAGE ON SEQUENCE package_name_id_seq TO evaluator;
+GRANT SELECT, USAGE ON SEQUENCE package_name_id_seq TO listener;
+GRANT SELECT, USAGE ON SEQUENCE package_name_id_seq TO vmaas_sync;
+
+CREATE TABLE IF NOT EXISTS package
+(
+    id          INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
+    -- TODO: inline vs extract ?
+    name_id     INT                                  NOT NULL REFERENCES package_name,
+    version     TEXT                                 NOT NULL CHECK (NOT empty(version)),
```
Josca, line 21: We can consider a separate table for version too. So we can add TODO here as well.

database_admin/migrations/025_packages.up.sql
```diff
@@ -0,0 +1,54 @@
+CREATE TABLE IF NOT EXISTS package_name
+(
+    id   INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
+    name TEXT                                 NOT NULL CHECK ( NOT empty(name)) UNIQUE
```
Josca, line 4: Some ` NOT empty(name)`s in this file have suffix space, some of them not, so let's unify it, please.

database_admin/migrations/025_packages.up.sql
```diff
@@ -0,0 +1,54 @@
...
+GRANT SELECT, USAGE ON SEQUENCE package_id_seq TO evaluator;
+GRANT SELECT, USAGE ON SEQUENCE package_id_seq TO listener;
+GRANT SELECT, USAGE ON SEQUENCE package_id_seq TO vmaas_sync;
+
+CREATE TABLE IF NOT EXISTS system_package
+(
+    system_id   INT NOT NULL REFERENCES system_platform,
+    package_id  INT NOT NULL REFERENCES package,
+    -- Use null to represent up-to-date packages
+    update_data JSONB DEFAULT NULL,
```
Josca, line 43: What do you want to store in this "update_data" JSONB? I think we don't need it. We store all update informations by pairing system-package in this table. Maybe we can add "current_version" column to store info about currently installed package version. And so we can add "package_version" enumerating table for that (also used in "package" table). Because this system_package pairing table will be probably huge.

database_admin/migrations/025_packages.up.sql
semtexzv: I only extracted name because it is expected to be duplicated, I don't think this is the same for version. In VMaaS we have separate table for EVR because it's duplicated across architectures

database_admin/migrations/025_packages.up.sql
semtexzv: Because if we store updates in separate table, the primary key for that table will be `system_id, package_id, index`. I thing for now it's better to do it this way, because of the size of that table, and issues when querying for data.

Josca: Why? As I know we need to store only the latest version for each package because only last update is required.

semtexzv: Yes, that is the current state, but in the future we might want to show all updates.

MichaelMraka: ok, it's all update version.

PR #288, opened 2020-07-16
Add package counts to system endpoints

semtexzv:
Adds pre-calculated package counts to `system_platform`.

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -135,7 +135,8 @@ func evaluateAndStore(tx *gorm.DB, system *models.SystemPlatform, vmaasData vmaa
 		evaluationCnt.WithLabelValues("error-system-pkgs").Inc()
 		return errors.Wrap(err, "Unable to update system packages")
 	}
-	err = updateSystemPlatform(tx, system, oldSystemAdvisories, newSystemAdvisories)
+	installed, updatable := getPackageCounts(vmaasData)
+	err = updateSystemPlatform(tx, system, oldSystemAdvisories, newSystemAdvisories, installed, updatable)
 	if err != nil {
 		evaluationCnt.WithLabelValues("error-update-system").Inc()
 		return errors.Wrap(err, "Unable to update system")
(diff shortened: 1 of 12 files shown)


evaluator/evaluate.go
```diff
@@ -233,7 +234,21 @@ func getPackages(tx *gorm.DB, data vmaas.UpdatesV2Response) (
 	return pkgByName, updatesByName, nil
 }
 
-func updateSystemPlatform(tx *gorm.DB, system *models.SystemPlatform, old, new SystemAdvisoryMap) error {
+func getPackageCounts(data vmaas.UpdatesV2Response) (int, int) {
+	installed := 0
```
Josca, line 238: So can we directly use installed := len(data.UpdateList) here and save periodic incrementing?

evaluator/evaluate.go
```diff
@@ -258,6 +273,8 @@ func updateSystemPlatform(tx *gorm.DB, system *models.SystemPlatform, old, new S
 	data["advisory_bug_count_cache"] = counts[2]
 	data["advisory_sec_count_cache"] = counts[3]
 	data["last_evaluation"] = time.Now()
+	data["packages_installed"] = installed
+	data["packages_updatable"] = updatable
```
Josca, line 277: I know there were some issues when updating "default values", e.g. 0 for int. Are you sure it will work here?

manager/controllers/systems.go
```diff
@@ -39,7 +39,9 @@ type SystemItemAttributes struct {
 	RheaCount      int        `json:"rhea_count" csv:"rhea_count" query:"system_platform.advisory_enh_count_cache"`
 	// Not used
 	//Enabled        bool       `json:"enabled" csv:"enabled" query:"(NOT system_platform.opt_out)"`
-	Stale bool `json:"stale" csv:"stale" query:"system_platform.stale"`
+	Stale     bool `json:"stale" csv:"stale" query:"system_platform.stale"`
```
Josca, line 42: These values are **not** included in SystemDetail response. It should be also checked in the systems_test.go and advisory_systems_test.go.

evaluator/evaluate.go
semtexzv: It should. We are already using it for advisory counts.

PR #294, opened 2020-07-27
Deduplicate strings in package data

semtexzv:
Use sha256 hashes for references, reducing the need to load primary keys from db


base/database/batch.go
```diff
@@ -63,7 +63,7 @@ func BulkInsertChunk(db *gorm.DB, objects interface{}, chunkSize int) []error {
 	}
 
 	if len(allErrors) > 0 {
-		return allErrors
+		return allErrors[0]
```
Josca, line 66: Why do you return only one error if there may be multiple ones?

database_admin/migrations/026_packages.up.sql
```diff
@@ -13,16 +13,25 @@ GRANT SELECT, USAGE ON SEQUENCE package_name_id_seq TO evaluator;
...
-    id          INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
+    id               INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
     -- TODO: inline vs extract ?
-    name_id     INT                                  NOT NULL REFERENCES package_name,
-    evra        TEXT                                 NOT NULL CHECK (NOT empty(evra)),
-    -- TODO: Inline vs extract to table ?
-    description TEXT                                 NOT NULL CHECK (NOT empty(description)),
-    -- TODO: Inline vs extract to table ?
-    summary     TEXT                                 NOT NULL CHECK (NOT empty(summary)),
+    name_id          INT                                  NOT NULL REFERENCES package_name,
```
Josca, line 31: Isn't "TODO" useless here now?

database_admin/schema/create_schema.sql
```diff
@@ -539,16 +539,22 @@ CREATE TABLE IF NOT EXISTS package_name
...
-    id          INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
+    id               INT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
     -- TODO: inline vs extract ?
-    name_id     INT                                  NOT NULL REFERENCES package_name,
-    evra        TEXT                                 NOT NULL CHECK (NOT empty(evra)),
-    -- TODO: Inline vs extract to table ?
-    description TEXT                                 NOT NULL CHECK (NOT empty(description)),
-    -- TODO: Inline vs extract to table ?
-    summary     TEXT                                 NOT NULL CHECK (NOT empty(summary)),
+    name_id          INT                                  NOT NULL REFERENCES package_name,
```
Josca, line 554: same here

vmaas_sync/pkg_sync.go
```diff
@@ -5,12 +5,15 @@ import (
...
 	"app/base/models"
 	"app/base/utils"
+	"crypto/sha256"
 	"github.com/RedHatInsights/patchman-clients/vmaas"
 	"github.com/antihax/optional"
 	"github.com/jinzhu/gorm"
 	"github.com/pkg/errors"
 )
 
+const chunkSize = 10 * 1024
```
Josca, line 15: I am not sure whether we use that convention everywhere, but I would prefer CHUNK_SIZE.

database_admin/migrations/026_packages.up.sql
Josca: If it's for easy pkg-name-based queries, we can comment it here, or in go struct definition and we can remove this TODO because it's no longer a question.

base/database/batch.go
semtexzv: We don't use the errors anywhere in the codebase, we always just propagate the first one.

vmaas_sync/pkg_sync.go
semtexzv: It's a private constant.

PR #296, opened 2020-07-28
Fix multiple installed packages handling

semtexzv:
- Old implementation still assumed one entry per package name.
- There was a bug in formatting NEVRAs

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -125,12 +125,12 @@ func evaluateAndStore(tx *gorm.DB, system *models.SystemPlatform, vmaasData vmaa
 		evaluationCnt.WithLabelValues("error-store-advisories").Inc()
 		return errors.Wrap(err, "Unable to store advisory data")
 	}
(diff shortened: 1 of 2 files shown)


evaluator/evaluate.go
```diff
@@ -164,25 +159,45 @@ func updateSystemPackages(tx *gorm.DB, system *models.SystemPlatform,
...
 	toStore := make([]models.SystemPackage, 0, len(updates))
-	for name, updateData := range updates {
+	for nevraStr, updateData := range updates {
 		var pkgUpdates []models.PackageUpdate
-
-		for _, upData := range updateData.UpdateData.AvailableUpdates {
+		// Parse each NEVRA in the input
+		nevra, err := utils.ParseNevra(nevraStr)
+		if err != nil {
+			utils.Log("nevra", nevraStr).Warn("Invalid nevra")
```
Josca, line 167: Would be good to count these invalid states, I know we can do it in another task.

PR #298, opened 2020-07-29
Small fixes, package data bugs.

semtexzv:
Fix issues observed with package data in CI.
- We have parsed NEVRAs incorrectly
- `delete_system` sql function was not updated.

### DIFF
--- a/database_admin/migrations/028_update_delete.up.sql
+++ b/database_admin/migrations/028_update_delete.up.sql
@@ -1,0 +1,40 @@
+CREATE OR REPLACE FUNCTION delete_system(inventory_id_in varchar)
+    RETURNS TABLE
+            (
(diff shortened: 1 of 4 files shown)


database_admin/migrations/028_update_delete.up.sql
```diff
@@ -0,0 +1,40 @@
+CREATE OR REPLACE FUNCTION delete_system(inventory_id_in varchar)
```
Josca, line 1: we should need to add more tests for whole this package-analysis related logic. I am afraid of other bugs that will be found later without better coverage

semtexzv: We have average coverage, the problem is we have only example data. Real data from vmaas is different.

PR #299, opened 2020-07-30
Package data perf improvements

semtexzv:
- Changed how we query for packages during evaluation
- Added prometheus metrics for additional evaluation steps

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -125,7 +125,7 @@ func evaluateAndStore(tx *gorm.DB, system *models.SystemPlatform, vmaasData vmaa
 		evaluationCnt.WithLabelValues("error-store-advisories").Inc()
 		return errors.Wrap(err, "Unable to store advisory data")
 	}
-	pkgByName, err := getPackages(tx, vmaasData)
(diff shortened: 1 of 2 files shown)


evaluator/evaluate.go
```diff
@@ -211,23 +212,28 @@ type namedPackage struct {
...
 	}
-
+	// Might return more data than we need (one EVRA being applicable to more packages
+	// But it was only way to get somewhat fast query plan which only uses index scans
 	var packages []namedPackage
 	err := tx.Table("package").
 		Select("concat(pn.name, '-', package.evra) nevra, package.*").
 		Joins("join package_name pn on package.name_id = pn.id").
-		Where("concat(pn.name, '-', package.evra) in (?)", pkgNevras).Find(&packages).Error
+		Where("pn.name in (?)", names).
```
Josca, line 235: Maybe it would be better not to extract package name into the separate table but rather keep it as string column in the package table and then create multi-column index on package.name and package.evra, what do you think?

semtexzv: We still have same problem, we need to search for specific name-evra pairs, and we can't do that with current library without performing a full table scan.

PR #301, opened 2020-08-13
fix (docs): openapi specs corrected

AlesKas:


docs/openapi.json
```diff
@@ -742,6 +742,7 @@
                 ],
                 "responses": {
                     "200": {
+                        "description": "Deleted",
```
Josca, line 745: I see different changes after running `./scripts/generate_docs.sh`. Did you do this update manually? Shouldn't it be updated in handlers annotations in go sources and than generated using the script? I expect that once we have openapi.json up to date I will get no changes after running that script. @semtexzv is it so?

semtexzv: Yes, spec files are generated from source code annotations.

AlesKas: Yes, I see it too, working on it.

PR #307, opened 2020-08-18
Add status endpoint

semtexzv:


manager/controllers/status.go
```diff
@@ -0,0 +1,14 @@
+package controllers
+
+import (
+	"app/base/database"
+	"github.com/gin-gonic/gin"
+)
+
+func Status(c *gin.Context) {
+	if err := database.Db.DB().Ping(); err != nil {
+		LogAndRespStatusError(c, 503, err, "Database not connected")
```
Josca, line 10: I would prefer to use constants for these status codes, as used in other controllers. We can replace constants also in other places (package_*) files.

semtexzv: Fixed.

PR #312, opened 2020-08-24
[WIP] Basic dynamic implementation of packages endpoint

semtexzv:
First implementation of per-account package data, currently calculating everything dynamically. If performance will be a problem, we'll move to pre-calculated table.


Josca (approved):
lgtm

PR #313, opened 2020-08-26
Update flags, add flag for project cyndi provided tags

semtexzv:


manager/controllers/utils.go
```diff
@@ -207,6 +207,9 @@ func ApplySearch(c *gin.Context, tx *gorm.DB, searchColumns ...string) *gorm.DB
 
 // Filter systems by tags,
 func ApplyTagsFilter(c *gin.Context, tx *gorm.DB, systemIDExpr string) (*gorm.DB, bool) {
+	if !utils.GetBoolEnvOrDefault("ENABLE_CYNDI_TAGS", false) {
```
Josca, line 210: Why to read it for each calling? Read it only once at the start and then only use bool var.

Josca (changes requested):

PR #314, opened 2020-08-26
added support for custom rbac roles

Josca:
related to rbac-config patch update: https://github.com/RedHatInsights/rbac-config/pull/84


semtexzv (approved):

semtexzv:
We must synchronize pushing these changes to changes in RBAC config.

Josca:
@semtexzv I don't think we need to synchronize it anyhow. Even if we can not expect "all:all" from RBAC service we still accept already supported "*:*". Actually shouldn't I leave `"*":"*"` row in patch rbac-config?

PR #315, opened 2020-08-27
feat (vmaas_sync): Cyndi metrics

AlesKas:


vmaas_sync/metrics.go
```diff
@@ -86,11 +87,24 @@ var (
...
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "cyndi_data",
+	}, []string{"type"})
+
+	cyndiCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "cyndi_systems",
+	}, []string{"last_upload"})
```
Josca, line 101: We can use only one lable `type` here. We can use it for "last_upload" flags (last1D, last7D...) and also for "tags_used" flag (count of systems with at least one tag), "sap_system" (count of systems with sap_system=true).

vmaas_sync/metrics.go
```diff
@@ -257,3 +273,111 @@ func updateProcessesMetrics() {
...
+func getCyndiData() (stats InventoryHosts, err error) {
+	err = database.Db.Table("inventory.hosts").Scan(&stats).Error
+	if err != nil {
+		return stats, errors.Wrap(err, "unable to get inventory hosts from db")
+	}
+	return stats, nil
+}
+
+func updateCyndiSystemMetrics() {
+	counts, err := getCyndiCounts(time.Now())
```
Josca, line 346: We should count with the case that inventory.hosts view is not available, as now in prod. So we should create it optional and enable/disable it using env variable `ENABLE_CYNDI_METRICS`.

vmaas_sync/metrics.go
```diff
@@ -257,3 +273,111 @@ func updateProcessesMetrics() {
...
+			cyndiStats.WithLabelValues(item.DisplayName + "-maxTag").Set(float64(0))
+			cyndiStats.WithLabelValues(item.DisplayName + "-minTag").Set(float64(0))
+		}
+		if systemProfile.SapSystem {
+			systemsSapSystem++
+		}
+	}
+	cyndiStats.WithLabelValues("systemsSapSystem").Set(float64(systemsSapSystem))
+	cyndiStats.WithLabelValues("systemsWithTags").Set(float64(systemsWithTag))
+	cyndiStats.WithLabelValues("systemsWithoutTags").Set(float64(systemsWithoutTag))
```
Josca, line 332: Sorry for that, use rather "all" label and then we can compute "without tags", "without sap" etc from that knowing only positive case of specific criteria.

vmaas_sync/metrics.go
```diff
@@ -257,3 +273,111 @@ func updateProcessesMetrics() {
...
+		} else {
+			systemsWithoutTag++
+			cyndiStats.WithLabelValues(item.DisplayName + "-maxTag").Set(float64(0))
+			cyndiStats.WithLabelValues(item.DisplayName + "-minTag").Set(float64(0))
+		}
+		if systemProfile.SapSystem {
+			systemsSapSystem++
+		}
+	}
+	cyndiStats.WithLabelValues("systemsSapSystem").Set(float64(systemsSapSystem))
```
Josca, line 330: Use cyndiSystemsCnt for that.

vmaas_sync/metrics.go
```diff
@@ -86,11 +87,24 @@ var (
...
 		Name:      "backend_processes",
 	}, []string{"state"})
+
+	cyndiStats = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "cyndi_data",
+	}, []string{"type"})
+
+	cyndiCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
Josca, line 97: Var name should be `cyndiSystemsCnt`.

vmaas_sync/metrics.go
```diff
@@ -86,11 +87,24 @@ var (
 		Subsystem: "vmaas_sync",
 		Name:      "backend_processes",
 	}, []string{"state"})
+
+	cyndiStats = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
Josca, line 91: I would use only `cyndiSystemsCnt` to measure systems count with specific properties (e.g. systems with some tags) and `cyndiTagsCnt` for tags counts (e.g. "maxPerSystem", "minPerSystem", cond. "avgPerSystem"). It would be also interesting to have count of unique tags used across all systems.

Josca (changes requested):
Good, but some updates are needed.

vmaas_sync/metrics.go
```diff
@@ -257,3 +273,111 @@ func updateProcessesMetrics() {
...
+		cyndiCnt.WithLabelValues(labels.LastUpload).Set(float64(count))
+	}
+}
+
+type cyndiCntLabels struct {
+	LastUpload string
+}
+
+func getCyndiCounts(refTime time.Time) (map[cyndiCntLabels]int, error) {
+	systemsQuery := database.Db.Model(&models.InventoryHosts{})
```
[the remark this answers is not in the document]
AlesKas: This line is same with `getSystemCount`, I only changed `SystemPlatform` to `InventoryHosts`.

conf/vmaas_sync.env
```diff
@@ -4,3 +4,5 @@ DB_PASSWD=vmaas_sync
 EVAL_TOPIC=patchman.evaluator.recalc
 
 ENABLE_REPO_BASED_RE_EVALUATION=true
+
+ENABLE_CYNDI_METRICS=false
```
semtexzv, line 8: Newline at the end of file

vmaas_sync/metrics.go
```diff
@@ -120,6 +123,12 @@ func update() {
 	updateAdvisoryMetrics()
 	updateSystemAdvisoriesStats()
 	updateProcessesMetrics()
+	enableCyndiMetrics, err := strconv.ParseBool(os.Getenv("ENABLE_CYNDI_METRICS"))
```
semtexzv, line 126: You can use `utils.GetBoolEnvOrDefault` to skip errors & should probably get this value at the start of the program by making it a module level variable

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,165 @@
...
+	}, []string{"type"})
+)
+
+type InventoryHosts []struct {
+	DisplayName   string
+	Tags          string
+	SystemProfile string
+}
+
+type Tags []struct {
```
semtexzv, line 36: It'd be much nicer to have a `Tag` struct instead of `Tags` and use slice elsewhere. Also these data- definition structs are usually stored in `base/models.go` so that we can reuse them for other purposes.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+	count["uniqueTagsCount"] = len(unique(uniqueTags))
+	count["systemsSapSystem"] = systemsSapSystem
+	count["systemsWithTag"] = systemsWithTag
+
+	return count, nil
+}
+
+func unique(inputTags []Tags) []Tags {
```
semtexzv, line 107: The convention is to have a `map[Tag]bool` and insert all values into this map, then create a new slice from the map.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+//Loads display_name, tags and system_profile from inventory.hosts
+//and loads them in InventoryHosts struct.
+//Tags are parsed into from jsonb string to TagSlice struct, which contains struct Tag with individual tags.
+//All tags are then stored in uniqueTagsSlice variable to compute, which tags are unique
+func getCyndiData() (map[string]int, error) {
+	count := map[string]int{}
+	uniqueTagsSlice := []models.TagSlice{}
```
semtexzv, line 57: This is in effect `[][]models.Tag`. is this the intention ? If not you can just use `models.TagSlice` and `append(uniqueTagsSlice,  tags...) down the line`

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,163 @@
...
+
+//Adds all tags to map, and checks, if any of these tags is unique
+//Uses method contains, which returns true, if tag is already in map's keys
+//Therefore value returned from method must be negated
+func unique(inputTags []models.TagSlice) map[models.Tag]bool {
+	mapOfUniqueTags := make(map[models.Tag]bool)
+
+	for _, singleTagSlice := range inputTags {
+		for _, singleTag := range singleTagSlice {
+			mapOfUniqueTags[singleTag] = !contains(mapOfUniqueTags, singleTag)
```
semtexzv, line 109: You don't have to use reflection here. The expression `mapOfUniqueTags[tag]` will return false if `tag` is not contained in the map

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+	mapTagsToUnique := make(map[models.Tag]bool)
+	var stats InventoryHosts
+	err := database.Db.Table("inventory.hosts").Scan(&stats).Error
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get inventory hosts from db")
+	}
+
+	var systemsWithTag int = 0
+	var systemsSapSystem int = 0
+	count["allSystemCount"] = len(stats)
```
semtexzv, line 65: Could be const

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, item := range stats {
+		var tags models.TagSlice
+		err := json.Unmarshal([]byte(item.Tags), &tags)
+		if err != nil {
+			utils.Log("err", err.Error()).Error("Not able to unmarshall")
```
semtexzv, line 70: You could skip the rest of the loop for this iteration or return error from function

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+	count["uniqueTagsCount"] = uniqueTags
+	count["systemsSapSystem"] = systemsSapSystem
+	count["systemsWithTag"] = systemsWithTag
+
+	return count, nil
+}
+
+func isUnique(uniqueTags map[models.Tag]bool, tagToCheck models.Tag) bool {
```
semtexzv, line 102: You don't need to check for uniqueness in a map. Assigning values in a loop and then iterating through map after the loop will do what you want

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		var systemProfile models.SystemProfile
+		err = json.Unmarshal([]byte(item.SystemProfile), &systemProfile)
+		if err != nil {
+			utils.Log("err", err.Error()).Error("Not able to unmarshall")
+		}
+		if tags != nil {
+			systemsWithTag++
+			maxTag := len(tags)
+			count[item.DisplayName] = maxTag
```
semtexzv, line 83: I'd use a histogram without using display name as a key

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		if systemProfile.SapSystem {
+			systemsSapSystem++
+		}
+	}
+	uniqueTags := 0
+	for _, value := range mapTagsToUnique {
```
semtexzv, line 90: `len(mapTagsToUnique)` will contain number of unique tags.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,145 @@
...
+	for labels, count := range counts {
+		cyndiSystemsCnt.WithLabelValues(labels.LastUpload).Set(float64(count))
+	}
+}
+
+type cyndiCntLabels struct {
+	LastUpload string
+}
+
+func getCyndiCounts(refTime time.Time) (map[cyndiCntLabels]int, error) {
```
semtexzv, line 123: `cyndiCntLabels` could be replaced by a string

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	cyndiTagsCnt.WithLabelValues(allSystemCount).Set(float64(stats.SystemsCount))
+	cyndiTagsCnt.WithLabelValues(uniqueTagsCount).Set(float64(stats.UniqueTags))
+	cyndiTagsCnt.WithLabelValues(systemsSapSystemCount).Set(float64(stats.SapCount))
+	cyndiTagsCnt.WithLabelValues(systemsWithTagsCount).Set(float64(stats.TagsCount))
+}
+
+func getCyndiData() (stats InventoryHosts, err error) {
+	err = database.Db.Table("inventory.hosts").
+		Select("COUNT(*) as systems_count, (SELECT COUNT(*) FROM inventory.hosts WHERE" +
```
Josca, line 62: Oh, that's not to easy to read. Is it possible to use gorm style for that? I mean something like:
~~~go
var nSAPSystems int
err := database.Db.Table("inventory.hosts").Where("system_profile -> 'sap_system' = 'true'").Count(&nSAPSystems).Error
// ...
~~~
You don't need to do it using single query. Rather split it to keep it more clear.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return stats, nil
+}
+
+func updateCyndiTags() {
+	var counts []struct {
+		JSONArrayLength int
+	}
+
+	if err := database.Db.Table("inventory.hosts").
```
Josca, line 80: I would prefer to keep it in two lines (first err :=, second: if). Then it's easier to debug. We use this approach everywhere so far.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+		Name:      "cyndi_systems",
+	}, []string{"type"})
+
+	cyndiTagsCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "cyndi_tags_count",
+	}, []string{"type"})
+
+	tagsPerSystem = prometheus.NewHistogram(prometheus.HistogramOpts{
```
Josca, line 34: Which histogram bins are used here. Are they appropriate for our usage? I would be good to discuss with someone from inventory/cyndi which values we can expect here to choose good resolution.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func updateCyndiTags() {
+	var counts []struct {
+		JSONArrayLength int
+	}
+
+	if err := database.Db.Table("inventory.hosts").
+		Select("json_array_length(tags::json)").
+		Find(&counts).Error; err != nil {
```
Josca, line 82: Could you simply use Pluck to array of ints?
https://gorm.io/docs/advanced_query.html#Pluck

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+		return
+	}
+
+	for labels, count := range counts {
+		cyndiSystemsCnt.WithLabelValues(labels).Set(float64(count))
+	}
+}
+
+func getCyndiCounts(refTime time.Time) (map[string]int, error) {
+	systemsQuery := database.Db.Model(&models.InventoryHosts{})
```
Josca, line 104: As this the only place of `models.InventoryHosts` struct usage, I think you can use `Table("inventory.hosts")` instead and not to define new struct at all. It's not needed if you don't use database row to go object mapping here.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,125 @@
...
+	}, []string{"type"})
+
+	tagsPerSystem = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "tags_per_system",
+	})
+)
+
+type InventoryHosts struct {
```
Josca, line 41: InventoryHostsStats is better struct name here. Moreover you use this name in models.go as well but for something else.

Josca (changes requested):
Ok, that's much better, but still something to improve imho.

vmaas_sync/metrics_cyndi.go
Josca: How is this resolved?

vmaas_sync/metrics_cyndi.go
Josca: Ok, we can remove this metric for now and add it later. Let's get this one done asap.

base/models/models.go
```diff
@@ -215,3 +215,32 @@ type TimestampKV struct {
 func (TimestampKV) TableName() string {
 	return "timestamp_kv"
 }
+
```
Josca, line 218: Try to avoid adding new models if we don't really need it.

base/models/models.go
```diff
@@ -215,3 +215,32 @@ type TimestampKV struct {
...
+	CulledTimestamp       *time.Time
+	Tags                  postgres.Jsonb
+	SystemProfile         postgres.Jsonb
+}
+
+func (InventoryHosts) TableName() string {
+	return "inventory.hosts"
+}
+
+type Tag struct {
```
Josca, line 236: Are these structs used anywhere?

Josca (changes requested):

base/models/models.go
AlesKas: Not anymore, deleted.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,118 @@
...
+const (
+	allSystemCount        = "allSystem"
+	uniqueTagsCount       = "uniqueTags"
+	systemsSapSystemCount = "systemsSapSystem"
+	systemsWithTagsCount  = "systemsWithTag"
+)
+
+var (
+	cyndiSystemsCnt = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "patchman_engine",
```
Josca, line 22: In on of previous PRs I added `Help` param to all metrics to describe their meaning. Please use it also for these new ones.

vmaas_sync/metrics_cyndi.go
Josca: Please name it InventoryHostsStats. It's more descriptive name for it.

vmaas_sync/metrics_cyndi.go
```diff
@@ -0,0 +1,118 @@
...
+
+func getCyndiData() (stats InventoryHosts, err error) {
+	err = database.Db.Table("inventory.hosts").Count(&stats.SystemsCount).Error
+	if err != nil {
+		utils.Log("err", err.Error()).Error("unable to update cyndi metrics")
+		return stats, err
+	}
+
+	err = database.Db.Table("(SELECT tags, cnt FROM (SELECT tags, COUNT(tags) over " +
+		"(partition by tags) cnt FROM inventory.hosts) as x where cnt = 1)  as unique_tags").
```
Josca, line 61: more spaces here

vmaas_sync/metrics_cyndi_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func TestCyndiTags(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	cyndiData, err := getCyndiData()
+	assert.Nil(t, err)
+	assert.Equal(t, 14, cyndiData.SystemsCount)
+	assert.Equal(t, 0, cyndiData.UniqueTags)
```
Josca, line 85: It's not too good testing with no tags? Please, add some testing ones and check counting here. Actually there should be some tags already. So why do we expect 0 unique tags here?

vmaas_sync/metrics.go
```diff
@@ -89,12 +89,20 @@ var (
...
 		Name:      "message_send_duration_seconds",
 	})
+
+	processesCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "patchman_engine",
+		Subsystem: "vmaas_sync",
+		Name:      "backend_processes",
+	}, []string{"state"})
+
+	enableCyndiMetrics = utils.GetBoolEnvOrDefault("ENABLE_CYNDI_METRICS", false)
```
Josca, line 99: Let's enable it by default. We are going to use it so we can only explicitly disable it when we really need.

conf/vmaas_sync.env
Josca, line 8: Set it rather to true.

database_admin/migrations/036_cyndi_permission.up.sql
```diff
@@ -0,0 +1 @@
+GRANT cyndi_reader TO vmaas_sync;
```
Josca, line 1: add empty line if possible, not the have this ugly red mark here.

vmaas_sync/metrics.go
```diff
@@ -89,12 +89,20 @@ var (
 		Subsystem: "vmaas_sync",
 		Name:      "message_send_duration_seconds",
 	})
+
+	processesCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
AlesKas: I had some conflicts, I didn't noticed, sorry, removed.

vmaas_sync/metrics_cyndi_test.go
AlesKas: There are only data that have same tags, so no tags are unique, I will add some data, with unique ones.

Josca:
@AlesKas do we really need
~~~
Merge confict

Merge conflict

Merge confict

Merge conflict
~~~
... in the commit message?

vmaas_sync/metrics_cyndi_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func TestCyndiTags(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	cyndiData, err := getCyndiData()
+	assert.Nil(t, err)
+	assert.Equal(t, 16, cyndiData.SystemsCount)
+	assert.Equal(t, 2, cyndiData.UniqueTags)
```
Josca, line 85: Why two unique tags? I see more unique tags in testing data:
~~~~
{"key": "k1", "value": "val1", "namespace": "ns1"}
{"key": "k2", "value": "val2", "namespace": "ns1"}
{"key": "k3", "value": "val4", "namespace": "ns1"}
{"key": "k3", "value": "val5", "namespace": "ns1"}
~~~~
.. so at least 4, if I count correctly.

vmaas_sync/metrics_cyndi_test.go
```diff
@@ -0,0 +1,88 @@
...
+func TestCyndiTags(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	cyndiData, err := getCyndiData()
+	assert.Nil(t, err)
+	assert.Equal(t, 16, cyndiData.SystemsCount)
+	assert.Equal(t, 3, cyndiData.UniqueTags)
+	assert.Equal(t, 14, cyndiData.SapCount)
+	assert.Equal(t, 16, cyndiData.TagsCount)
```
Josca, line 87: Rename the field to SystemsWithTags, please. And the number seems to be wrong.

Josca (changes requested):

Josca (commented):
Great, thanks.

PR #324, opened 2020-09-09
Detect stuck kafka readers

semtexzv:


semtexzv:
Temporarily fixed with more permanent fix coming with change of kafka implementation

PR #327, opened 2020-09-09
Change kafka implementation

semtexzv:
Changes kafka implementation to use new library. 

Do not merge, it should be beneficial to compare performance of evaluator before and after.


listener/listener_test.go
```diff
@@ -13,10 +13,9 @@ func TestRunReaders(t *testing.T) {
 	nReaders := 0
 	wg := sync.WaitGroup{}
 	wg.Add(1)
-	runReaders(&wg, mqueue.CreateCountedMockReader(&nReaders))
+	runReaders(mqueue.CreateCountedMockReader(&nReaders))
```
Josca, line 16: Do we still need wg here?

Josca (commented):
It looks good. But as you mentioned perf. testing... wouldn't it be useful to keep old implementation usable as well and allow switching between these two frameworks using env variable?

semtexzv:
Main goal is to simplify the application, leave a lot of partition handling to the library. I don't want to complicate things further by using 2 libraries at the same time

Josca:
@semtexzv ok, so just please process my "wg" note, I can approve it then.

Josca (approved):
Ok, I hope it will work. Thanks for big upgrade.

PR #329, opened 2020-09-10
Fix listener behavior

semtexzv:
- Process uploads on update events.
- Cleanup the listener code


Josca (approved):
Uh, I did not expect so many updates. But guessing it's code cleaning. Ok.

PR #336, opened 2020-09-15
Don't log system ids in vmaas sync

semtexzv:
Last recalc we spent 15 minutes just logging all of the systems and sending messages

### DIFF
--- a/vmaas_sync/send_messages.go
+++ b/vmaas_sync/send_messages.go
@@ -53,7 +53,6 @@ func sendMessages(ctx context.Context, inventoryIDs ...string) {
 	events := make([]mqueue.PlatformEvent, len(inventoryIDs))
 
 	for i, id := range inventoryIDs {
-		utils.Log("inventoryID", id).Debug("Sending evaluation kafka message")
 		events[i] = mqueue.PlatformEvent{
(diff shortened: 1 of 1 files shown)


Josca:
@semtexzv yes, but it's debug log. Wouldn't it better to switch to "INFO" log level in prod rather than removing it?

vmaas_sync/send_messages.go
```diff
@@ -53,7 +53,6 @@ func sendMessages(ctx context.Context, inventoryIDs ...string) {
 	events := make([]mqueue.PlatformEvent, len(inventoryIDs))
 
 	for i, id := range inventoryIDs {
-		utils.Log("inventoryID", id).Debug("Sending evaluation kafka message")
```
Josca, line 56: Are you sure the long duration was caused by logging? It could be caused by sending messages through the Kafka. Anyway we can set INFO logging level in prod and so this message will be disabled.

Josca (changes requested):

vmaas_sync/send_messages.go
Josca: Ok, merging. We can simply revert it later if we need.

PR #337, opened 2020-09-16
Added metric counter for reporter type

Josca:


listener/listener.go
```diff
@@ -22,6 +24,16 @@ func configure() {
 	evalTopic := utils.GetenvOrFail("EVAL_TOPIC")
 
 	evalWriter = mqueue.WriterFromEnv(evalTopic)
+
+	parseValidReporters()
+}
+
+func parseValidReporters() {
```
semtexzv, line 31: Can't we get the DB PR merged and use database to store list of valid reporters ? This duplicates the number of places, where we store list of valid reporters.

Josca: Yes, it's done in the second PR.

semtexzv: But that PR keeps the `parseValidReporters` method. It should then remove it, so that  there is only 1 source of this data.

Josca: Yes, but it loads it from db: https://github.com/RedHatInsights/patchman-engine/pull/338/files so there IS only one source of reporters set.

semtexzv (approved):

PR #341, opened 2020-09-21
feat (manager): endpoint to return single package

AlesKas:


manager/controllers/package_detail.go
```diff
@@ -0,0 +1,118 @@
...
+	})
+}
+
+// @Summary Show me metadata of selected package
+// @Description Show me metadata of selected package
+// @ID LatestPackage
+// @Security RhIdentity
+// @Accept   json
+// @Produce  json
+// @Param    package_name    path    string   true "Package name"
```
Josca, line 102: I think it should be package_param here. And please provide more info how the param should look like, there are some possibilities of usage (package_name - latest, nevra - exact version).

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func packageEvraHandler(c *gin.Context, nevra *utils.Nevra) {
+	packageName := nevra.Name
+	evra := nevra.EVRAString()
+
+	query := singlePackageEvraQuery(packageName, evra)
+
+	var packageInfo models.Package
+	err := query.Find(&packageInfo).Error
```
Josca, line 75: What if it's not found?

manager/controllers/package_detail_test.go
```diff
@@ -0,0 +1,44 @@
...
+	assert.Equal(t, http.StatusOK, w.Code)
+	var output PackageDetailResponse
+	assert.Greater(t, len(w.Body.Bytes()), 0)
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+	assert.Equal(t, "kernel", output.Name)
+	assert.Equal(t, "RH-7", output.Data.Advisory.Name)
+	assert.Equal(t, "5.6.13-201.fc31.x86_64", output.Data.Package.EVRA)
+}
+
+func TestEvraPackage(t *testing.T) {
```
Josca, line 29: Please add also some edge cases (non-existing package_name, non-existing nevra, existing package_name with non-existing version). Please ensure to return correct status codes. See `AdvisoryDetailHandler` and use similar way.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,118 @@
...
+	return advisory, err
+}
+
+type PackageDetailResponse struct {
+	Data SinglePackage `json:"data"`
+	Name string        `json:"name"`
+}
+
+type SinglePackage struct {
+	Package  models.Package          `json:"package"`
```
Josca, line 38: Please define special structs for response, PackageDetailItem, PackageDetailAttributes. Everything should be analogical to advisory_detail.go, do it consistently.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,118 @@
...
+// @Summary Show me metadata of selected package
+// @Description Show me metadata of selected package
+// @ID LatestPackage
+// @Security RhIdentity
+// @Accept   json
+// @Produce  json
+// @Param    package_name    path    string   true "Package name"
+// @Success 200 {object} PackageDetailResponse
+// @Router /api/patch/v1/packages/{package_param} [get]
+func PackageRouter(c *gin.Context) {
```
Josca, line 105: This should be PackageDetailHandler.

docs/openapi.json
```diff
@@ -1468,6 +1525,79 @@
...
+                    "evra": {
+                        "type": "string"
+                    },
+                    "id": {
+                        "type": "integer"
+                    },
+                    "nameID": {
+                        "type": "integer"
+                    },
+                    "summaryHash": {
```
Josca, line 1593: Hashes are useless for endpoint caller. He needs summary or description itself. Hashes are used only internally for pairing.

Josca (changes requested):
Good, we are almost there but still something to do.

manager/controllers/package_detail.go
AlesKas: Then 500 is returned with message that specified package was not found.

Josca: 5xx status codes should be returned only on error cases what's not on "Not found". It should return 404 status in that case. Again see the AdvisoryDetailHandler and implement is similarly.

AlesKas: Noticed, returning 404.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,124 @@
...
+		Where("package.evra = ?", evra)
+}
+
+func getAdvisoryByID(advID int) (models.AdvisoryMetadata, error) {
+	var advisory models.AdvisoryMetadata
+	err := database.Db.Where("id = ?", advID).First(&advisory).Error
+	return advisory, err
+}
+
+type PackageDetailResponse struct {
```
Josca, line 32: Please make the structures consistent with AdvisoryDetailResponse and SystemDetailResponse. It's still different.
~~~go
type PackageDetailResponse struct {
	Data PackageDetailItem `json:"data"`
}

type PackageDetailItem struct {
	Attributes PackageDetailAttributes  `json:"attributes"`
	ID         string                   `json:"id"` // package nevra
	Type       string                   `json:"type"` // this will be always "package"
}

type PackageDetailAttributes struct {
	Description  string   `json:"description"` // not hashes for des. and sum. but content itself
	Summary      string   `json:"summary"`
	Name         string   `json:"name"` // package name
	Version      string   `json:"version"` // put whole evra here for now
	AdvisoryID   string   `json:"advisory_id"` // e.g. RHBA-2020:1234
	// it seems that's all we have now, maybe we will split version and arch later
}
~~~
@semtexzv @MichaelMraka please, correct me if I am wrong.

docs/openapi.json
```diff
@@ -1468,6 +1542,50 @@
                     }
                 }
             },
+            "models.AdvisoryMetadata": {
```
Josca, line 1545: There is some mistake this object should not be used in API it's model for object-relation mapping.

Josca (changes requested):
next update, thanks

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,139 @@
...
+type PackageDetailAttributes struct {
+	Description string `json:"description"`
+	Summary     string `json:"summary"`
+	Name        string `json:"name"`
+	Version     string `json:"version"`
+	AdvisoryID  string `json:"advisory_id"`
+}
+
+func packageLatestHandler(c *gin.Context, packageName string) {
+	query := singlePackageLatestQuery(packageName)
```
Josca, line 51: To be correct, firstly we should check whether input is valid package_name, using some regex. If not we should return Bad request instead of directly trying to search for any input string.

Josca (changes requested):
almost there, some small updates required

Josca:
@AlesKas btw, there is a typo in PR title.

manager/controllers/package_detail.go
AlesKas: This should be done in `PackageDetailHandler` function, which should determine what function should be called, so handlers should not care about input.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,193 @@
...
+	"app/base/database"
+	"app/base/models"
+	"app/base/utils"
+	"github.com/gin-gonic/gin"
+	"github.com/jinzhu/gorm"
+	"net/http"
+)
+
+func singlePackageLatestQuery(pkgName string) *gorm.DB {
+	return database.Db.Table("package pkg").Select("*").Joins("JOIN package_name pn ON pkg.name_id = pn.id").
```
Josca, line 13: Please, put the Joins on the next line.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,193 @@
...
+// @Param    package_name    path    string   true "package_name - latest, nevra - exact version"
+// @Success 200 {object} PackageDetailResponse
+// @Router /api/patch/v1/packages/{package_name} [get]
+func PackageDetailHandler(c *gin.Context) {
+	parameter := c.Param("package_name")
+	if parameter == "" {
+		c.JSON(http.StatusBadRequest, utils.ErrorResponse{Error: "package_param not found"})
+		return
+	}
+
```
Josca, line 186: To be more correct we should add parameter validation here. Nevra should include only specific characters. We should return "BadRequest" if e.g. % is included. I think only `a-zA-Z0-9\+-\._^~` are allowed for nevra. But of course it's not critical.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,193 @@
...
+	summary, err := getPackageStrings(packageInfo.SummaryHash)
+	if gorm.IsRecordNotFoundError(err) {
+		LogAndRespNotFound(c, err, "summary not found")
+		return
+	}
+
+	descr, err := getPackageStrings(packageInfo.DescriptionHash)
+	if gorm.IsRecordNotFoundError(err) {
```
Josca, line 151: Using correct sql query and joins you can load all info once into the `PackageDetailAttributes` object. It will be more efficient and shorter. Try it.
Also you can make basic part of query for both variants using the same method (joint package, package_name, advisory, strings). Then you should apply other criteria depending on whether it's latest (order by advisory pub.date) or exact version (where version is ...).

manager/controllers/package_detail_test.go
```diff
@@ -0,0 +1,83 @@
...
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/", nil)
+	core.InitRouter(PackageDetailHandler).ServeHTTP(w, req)
+	assert.Equal(t, http.StatusBadRequest, w.Code)
+	var errResp utils.ErrorResponse
+	ParseReponseBody(t, w.Body.Bytes(), &errResp)
+	assert.Equal(t, "package_param not found", errResp.Error)
+}
```
Josca, line 83: Great, now it seems we have all possible cases covered. :+1:

manager/controllers/package_detail_test.go
```diff
@@ -0,0 +1,83 @@
...
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/packages/kernel", nil)
+	core.InitRouterWithParams(PackageDetailHandler, "3", "GET", "/packages/:package_name").
+		ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+	var output PackageDetailResponse
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+	assert.Equal(t, "kernel", output.Data.Attributes.Name)
+	assert.Equal(t, "RH-7", output.Data.Attributes.AdvisoryID)
+	assert.Equal(t, "5.6.13-201.fc31.x86_64", output.Data.Attributes.Version)
```
Josca, line 25: Please rather check all result fields (desc., summ, ID), thanks. Also for ExactPackage test.

Josca (changes requested):
Still something, sorry.

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+}
+
+func packageEvraHandler(c *gin.Context, nevra *utils.Nevra) {
+	if !packageNameIsValid(nevra.Name) {
+		c.JSON(http.StatusBadRequest, utils.ErrorResponse{Error: "invalid package name"})
+		return
+	}
+
+	packageName := nevra.Name
```
Josca, line 96: do we need this row?

manager/controllers/package_detail.go
```diff
@@ -0,0 +1,137 @@
...
+	return database.Db.Table("package pkg").Select("*").
+		Joins("JOIN package_name pn ON pkg.name_id = pn.id").
+		Joins("JOIN (SELECT id, name as adv_id, public_date FROM advisory_metadata) as adv ON pkg.advisory_id = adv.id").
+		Joins("JOIN (SELECT id, value as description FROM strings) as "+
+			"string_descr ON pkg.description_hash = string_descr.id").
+		Joins("JOIN (SELECT id, value as summary FROM strings) as string_sum ON pkg.summary_hash = string_sum.id").
+		Where("pn.name = ?", pkgName)
+}
+
+// func singlePackageEvraQuery(pkgName string, evra string) *gorm.DB {
```
Josca, line 22: Please remove this commented block. It's useless here.

Josca (changes requested):
Great, it looks like it works. Just two cosmetic updates and it's done.

Josca (approved):
Ok, that looks good. Let's push it forward.

PR #348, opened 2020-09-28
Add summary to package data

semtexzv:


manager/controllers/packages.go
```diff
@@ -30,15 +31,15 @@ type PackagesResponse struct {
 }
 
 func packagesQuery(acc string) *gorm.DB {
-	return database.Db.
+	return database.Db.Debug().
```
Josca, line 34: It's forgotten Debug() I guess

manager/controllers/packages.go
```diff
@@ -16,11 +16,12 @@ var PackagesOpts = ListOpts{
 	DefaultSort:    "name",
 }
 
+// nolint: lll
 type PackageItem struct {
 	Name             string `json:"name" query:"pn.name"`
 	SystemsInstalled int    `json:"systems_installed" query:"count(sp.id)"`
 	SystemsUpdatable int    `json:"systems_updatable" query:"count(sp.id) filter (where spkg.update_data is not null)"`
-	// Description      string `json:"description"`
+	Summary          string `json:"summary" query:"(select s.value from package p inner join strings s on p.summary_hash = s.id inner join advisory_metadata am on p.advisory_id = am.id where p.name_id = pn.id order by am.public_date limit 1)"`
```
Josca, line 24: Funny solution, ok it's working. Is it at least possible to split the query text into multiple lines to be more clear?

Josca (changes requested):

manager/controllers/packages.go
semtexzv: Tried, however it wasn't accepted by the parser for `query` attribute

Josca:
pipeline failed

Josca (approved):

PR #349, opened 2020-09-28
Inventory-like workloads filtering

semtexzv:
Implemented using new nested query parsing functionality


manager/controllers/systems_test.go
```diff
@@ -136,6 +136,22 @@ func TestSystemsTags(t *testing.T) {
 	assert.Equal(t, "00000000-0000-0000-0000-000000000001", output.Data[0].ID)
 }
 
+func TestSystemsWorkloads(t *testing.T) {
```
Josca, line 139: I would prefer more tests here. What about some edge cases? Does it work for sap_system=false filter? Does it works for empty "sap_sids" input?

Josca (changes requested):
just one thing, thanks

manager/controllers/systems_test.go
semtexzv: Added

Josca (approved):
thanks

PR #351, opened 2020-09-29
Partition system_package table by account

semtexzv:


database_admin/migrations/035_partition_system_package.up.sql
```diff
@@ -0,0 +1,41 @@
...
+                    ' FOR VALUES WITH ' || ' ( MODULUS ' || text(parts) || ', REMAINDER ' || text(I) || ')' ||
+                    rest || ';';
+            I = I + 1;
+        END LOOP;
+END;
+$$ LANGUAGE plpgsql;
+
+DROP TABLE IF EXISTS system_package;
+
+CREATE TABLE IF NOT EXISTS system_package
```
Josca, line 20: "IF NOT EXISTS" is useless here.

manager/controllers/packages.go
```diff
@@ -43,11 +43,19 @@ func packagesQuery(acc string) *gorm.DB {
 
 // @Summary Show me all installed packages across my systems
 // @Description Show me all installed packages across my systems
-// @ID packages
+// @ID listPackages
 // @Security RhIdentity
 // @Accept   json
 // @Produce  json
-// @Success 200 {object} PackageSystemsResponse
+// @Param    limit          query   int     false   "Limit for paging, set -1 to return all"
```
Josca, line 50: This is different thing, right? Wouldn't be better to put it to sepparate pr?

evaluator/evaluate.go
```diff
@@ -255,7 +257,8 @@ func loadPackages(tx *gorm.DB, systemID int, data vmaas.UpdatesV2Response) (map[
 	err := tx.Table("package").
 		Select("concat(pn.name, '-', package.evra) nevra, package.*, sp.*").
 		Joins("join package_name pn on package.name_id = pn.id").
-		Joins("left join system_package sp on sp.system_id = ? and sp.package_id = package.id", systemID).
+		// nolint: lll
+		Joins("left join system_package sp on sp.rh_account_id = ? AND  sp.system_id = ? AND sp.package_id = package.id", accountID, systemID).
```
Josca, line 261: better split it, using "``" for string definition. A bit more spaces then needed.

database_admin/schema/create_schema.sql
```diff
@@ -426,6 +426,23 @@ BEGIN
 END;
 $fun$ LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION create_table_partitions(tbl regclass, parts INTEGER, rest text)
```
Josca, line 429: Why the regclass type is used here?

Josca (commented):
Looks good to me. Just a few questions/notes. @MichaelMraka, please check it as well. Also I don't see where the hashing and assigning to partition is done for new system_platform items.

database_admin/schema/create_schema.sql
semtexzv: Only allows registered names like tables or indexes in the function.

manager/controllers/packages.go
semtexzv: Removed

manager/controllers/package_systems.go
```diff
@@ -25,7 +25,7 @@ func packageSystemsQuery(acc string, pkgName string) *gorm.DB {
 	return database.Db.
 		Table("system_platform").
 		Joins("inner join rh_account ra on system_platform.rh_account_id = ra.id").
-		Joins("inner join system_package spkg on spkg.system_id = system_platform.id").
+		Joins("inner join system_package spkg on  on spkg.rh_account_id = ra.id AND spkg.system_id = system_platform.id").
```
MichaelMraka: on  on spkg.rh_account_id = ra.id AND 
^^ duplicated keyword

evaluator/evaluate.go
semtexzv: Fixed by moving to `WHERE`

database_admin/migrations/035_partition_system_package.up.sql
MichaelMraka: @Josca Hm, this PR is changing faster than I can write review ;).

manager/controllers/package_systems.go
semtexzv: Fixed

MichaelMraka: I'd prefer to fix it in this commit not next one, just for the sake of clarity. ;)

evaluator/evaluate.go
MichaelMraka: Same here - I'd prefer to fix it in this commit not next one, just for the sake of clarity. ;)

Josca (approved):

MichaelMraka (approved):

PR #352, opened 2020-09-29
Use account ID instead of name in most queries

semtexzv:


manager/middlewares/authentication.go
```diff
@@ -1,13 +1,21 @@
...
+	"app/base/models"
 	"app/base/utils"
 	"github.com/gin-gonic/gin"
 	"net/http"
+	"sync"
 )
 
 const KeyAccount = "account"
 
+var AccountIDCache = struct {
```
Josca, line 14: Don't you rather use lib for cache object? https://github.com/hashicorp/golang-lru

manager/middlewares/authentication.go
```diff
@@ -22,12 +30,24 @@ func Authenticator() gin.HandlerFunc {
 			c.AbortWithStatusJSON(http.StatusUnauthorized, utils.ErrorResponse{Error: "Invalid x-rh-identity header"})
 			return
 		}
-		c.Set(KeyAccount, identity.Identity.AccountNumber)
+		AccountIDCache.Lock.Lock()
```
Josca, line 33: Is it needed to lock it for both read and write? Isn't locking for write enought?

manager/controllers/utils.go
```diff
@@ -150,6 +150,7 @@ func ExportListCommon(tx *gorm.DB, c *gin.Context, opts ListOpts) (*gorm.DB, err
 
 func ListCommon(tx *gorm.DB, c *gin.Context, path string, opts ListOpts) (*gorm.DB, *ListMeta, *Links, error) {
 	tx, limit, offset, sortFields, filters, err := checkBadRequest(tx, c, opts)
+
```
Josca, line 153: update by mistake I guess

Josca (changes requested):

manager/middlewares/authentication.go
semtexzv: No need, since cache is never cleaned up. I'd rather leave it as is, than pull in another dependency

manager/middlewares/authentication.go
semtexzv: Not sure, anyway, one mutex for short time in this piece of code will not affect performance.

Josca (approved):
lgtm

PR #357, opened 2020-10-01
Fix issue with evaluation

semtexzv:
We perform left join and in previous PR moved additional join params to where clase, which changed behavior of whole query.

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -143,12 +143,13 @@ func evaluateAndStore(tx *gorm.DB, system *models.SystemPlatform, vmaasData vmaa
 	return nil
 }
 
-func deleteOldSystemPackages(tx *gorm.DB, systemID int, packagesByNEVRA map[utils.Nevra]namedPackage) error {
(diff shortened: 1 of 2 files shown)


evaluator/evaluate.go
```diff
@@ -257,8 +258,9 @@ func loadPackages(tx *gorm.DB, accountID, systemID int,
 	err := tx.Table("package").
 		Select("concat(pn.name, '-', package.evra) nevra, package.*, sp.*").
 		Joins("join package_name pn on package.name_id = pn.id").
-		Joins(`left join system_package sp on sp.package_id = package.id`).
-		Where("sp.rh_account_id = ? AND sp.system_id = ?", accountID, systemID).
+		// We need to perform left join, so thats why the parameters are here
```
Josca, line 261: What to add some tests to reveal issues like this?

Josca (changes requested):

evaluator/evaluate.go
semtexzv: Added more checks to `TestEvaluate`

Josca: good, thanks

Josca (approved):

PR #364, opened 2020-10-06
Better check to skip writing package updates, reduce loaded columns

semtexzv:
Should improve performance by loading less data and overwriting less data.

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -7,6 +7,7 @@ import (
 	"app/base/models"
 	"app/base/mqueue"
 	"app/base/utils"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


evaluator/evaluate.go
```diff
@@ -308,11 +306,12 @@ func loadPackages(tx *gorm.DB, accountID, systemID int,
...
 		return nil, errors.Wrap(err, "loading packages")
 	}
+
 	pkgByNevra := map[utils.Nevra]namedPackage{}
 	for _, p := range packages {
-		nevra, err := utils.ParseNevra(p.NEVRA)
+		nevra, err := utils.ParseNameEVRA(p.Name, p.EVRA)
 		if err != nil {
-			utils.Log("err", err.Error(), "nevra", p.NEVRA).Warn("Unable to parse nevra")
+			utils.Log("err", err.Error(), "name", p.Name, p.EVRA).Warn("Unable to parse nevra")
```
Josca, line 314: There is odd count of Log() args. It have to be even (name - value pairs) number.

Josca (changes requested):

evaluator/evaluate.go
semtexzv: Thanks, fixed

PR #367, opened 2020-10-07
Skip evaluation if message is older than last_evaluation

semtexzv:


evaluator/evaluate.go
```diff
@@ -56,7 +56,7 @@ func configure() {
 	vmaasClient = vmaas.NewAPIClient(vmaasConfig)
 }
 
-func Evaluate(ctx context.Context, inventoryID string, evaluationType string) error {
+func Evaluate(ctx context.Context, inventoryID string, t *time.Time, evaluationType string) error {
```
Josca, line 59: What to name it better than `t`? E.g. "evalRequestTst" or something?

listener/upload.go
```diff
@@ -97,9 +97,10 @@ func HandleUpload(event HostEvent) error {
 			return nil
 		}
 	}
-
+	t := time.Now().Format(base.Rfc3339NoTz)
```
Josca, line 100: This should be done also (and mainly) in vmaas_sync when sending recalc messages.

Josca, line 100: Have you tried to use time.Time type for PlatformEvent.Timestamp field? Maybe it would ser/deserialize it out of the box.

evaluator/evaluate.go
Josca: oh, `tRequested` would be clear, I think.

Josca (changes requested):
Good, some notes and proposals introduced.

listener/upload.go
semtexzv: If the field is missing, then it will be deserialized to zero type - which is weird to work with in the case of `time.Time`. Pointer is much clearer

Josca: @semtexzv you can use `*time.Time` so it should be nil once field is not presented.

semtexzv: Done, thanks

semtexzv: Resolved by creating a wrapper which works with required datetime format.

Josca:
@semtexzv ok, just ensure passing pipeline and merge it

listener/upload.go
Josca: Ok, but I don't see it. Still see string ser/deser.

PR #371, opened 2020-10-09
Wrap mutex locking + unlocking in inline function

semtexzv:


manager/middlewares/authentication.go
```diff
@@ -30,19 +30,22 @@ func Authenticator() gin.HandlerFunc {
...
-		AccountIDCache.Lock.Lock()
-		defer AccountIDCache.Lock.Unlock()
-
-		if id, has := AccountIDCache.Values[identity.Identity.AccountNumber]; has {
-			c.Set(KeyAccount, id)
-		} else {
-			var acc models.RhAccount
-			if err := database.Db.Where("name = ?", identity.Identity.AccountNumber).Find(&acc).Error; err != nil {
-				c.AbortWithStatusJSON(http.StatusUnauthorized, utils.ErrorResponse{Error: "Could not find rh_account"})
+		func() {
```
Josca, line 33: Will it be really called? Have you tested that?

semtexzv: Moved to separate function

PR #372, opened 2020-10-09
Add privileged interface

semtexzv:
Use platform-go-middleware for parsing identity, Add a new api group accepting only associate auth, which comes from turnpike.

### DIFF
--- a/base/utils/identity.go
+++ b/base/utils/identity.go
@@ -3,26 +3,12 @@ package utils
 import (
 	"encoding/base64"
 	"encoding/json"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
-type Entitlement struct {
-	Entitled bool `json:"is_entitled"`
-}
+func ParseIdentity(identityString string) (*identity.Identity, error) {
+	var ident identity.XRHID
 
-type IdentityDetail struct {
-	AccountNumber string `json:"account_number"`
-	Type          string `json:"type"`
-	// Additional information, we don't parse this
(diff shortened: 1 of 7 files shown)


base/utils/identity.go
```diff
@@ -3,50 +3,20 @@ package utils
...
 
-type Identity struct {
-	Entitlements map[string]Entitlement `json:"entitlements"`
-	Identity     IdentityDetail         `json:"identity"`
-}
-
-func ParseIdentity(identityString string) (*Identity, error) {
-	var ident Identity
 	decoded, err := base64.StdEncoding.DecodeString(identityString)
+	print(string(decoded))
```
Josca, line 13: print?

manager/middlewares/authentication.go
```diff
@@ -43,17 +45,34 @@ func Authenticator() gin.HandlerFunc {
 		}
 		utils.Log("ident", identStr).Trace("Identity retrieved")
 
-		identity, err := utils.ParseIdentity(identStr)
+		ident, err := utils.ParseIdentity(identStr)
```
Josca, line 48: What's the purpose of this change?

Josca (changes requested):
looks good, I think I found one forgotten print, thanks

manager/middlewares/authentication.go
semtexzv: Conflict with package name

Josca (approved):

PR #373, opened 2020-10-12
Move syncing advisories into separate goroutine

semtexzv:
Should resolve vmaas_sync crashing issue

### DIFF
--- a/vmaas_sync/vmaas_sync.go
+++ b/vmaas_sync/vmaas_sync.go
@@ -90,21 +90,25 @@ func runWebsocket(conn *websocket.Conn, handler Handler) error {
 	}
 }
 
+func onRefreshed() {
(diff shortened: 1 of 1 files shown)


Josca (changes requested):
1) Why should this help?
2) Tests pipeline failed.

semtexzv:
It seems that connection to vmaas times out in prod

Josca:
So it does not solve the whole issue - advisories are not synced. It mill maybe just resolve issue of restarting, right?

Josca:
@semtexzv ok, makes sense, so just ensure passing pipeline and then feel free to merge it.

semtexzv:
Superseded by https://github.com/RedHatInsights/patchman-engine/pull/376

PR #379, opened 2020-10-13
use latest_only option for vmaas api

MichaelMraka:
to filter out multiple version of installonly packages

### DIFF
--- a/listener/upload.go
+++ b/listener/upload.go
@@ -346,6 +346,7 @@ func processUpload(account string, host *Host) (*models.SystemPlatform, error) {
 		ModulesList:    processModules(&systemProfile),
 		Basearch:       systemProfile.Arch,
 		SecurityOnly:   false,
+		LatestOnly:     true,
 	}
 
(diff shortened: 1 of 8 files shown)


listener/upload.go
```diff
@@ -355,6 +355,7 @@ func processUpload(account string, host *Host) (*models.SystemPlatform, error) {
 		ModulesList:    processModules(&systemProfile),
 		Basearch:       systemProfile.Arch,
 		SecurityOnly:   false,
+		LatestOnly:		true,
```
Josca, line 358: Wrong formating, will fail on "gofmt error". Anyway, you need to update new version of vmaas client firstly. It's failing on `listener/upload.go:358:3: unknown field 'LatestOnly' in struct literal of type vmaas.UpdatesV3Request`

Josca (changes requested):

listener/upload.go
MichaelMraka: fixed

Josca (approved):
cool, finally we've done it :+1:

PR #383, opened 2020-10-13
Add missing attribute to package systems

semtexzv:
We had missing attribute on this endpoint. Needed for package systems table.


Josca (commented):
can be merged after passed tests

PR #390, opened 2020-10-15
Fix missing non-updatable packages

semtexzv:


evaluator/evaluate.go
```diff
@@ -301,10 +308,9 @@ func loadPackages(tx *gorm.DB, accountID, systemID int,
 	// But it was only way to get somewhat fast query plan which only uses index scans
 	var packages []namedPackage
 	err := tx.Table("package").
-		Select("pn.name, package.id as package_id, package.evra, sp.update_data").
+		Select("pn.name, package.id as package_id, package.evra, (sp.system_id IS NOT NULL) as was_stored, sp.update_data").
```
Josca, line 311: Do we want to load all system packages for the given system and join it with package related info here? Why do we use "left join" here?

semtexzv: We had it this way before, but I wanted to reduce amount of loaded columns to make the evaluation faster. We basically use this to load package data and previous evaluation data in single step.

The problem was that when we loaded the update_data null, it could mean 2 things. Either the system_package entry was not stored, or it was stored, with `update_data` was null.

By adding the was_stored, we can resolve what case it was without doing 2 complex queries for evalutaion.

manager/controllers/system_packages.go
```diff
@@ -18,12 +18,12 @@ type SystemPackagesAttrs struct {
 	EVRA        string `json:"evra" query:"p.evra"`
 	Summary     string `json:"summary" query:"sum.value"`
 	Description string `json:"description" query:"descr.value"`
-	Updatable   bool   `json:"updatable" query:"(json_array_length(spkg.update_data::json) > 0)"`
+	Updatable   bool   `json:"updatable" query:"spkg.update_data IS NOT NULL"`
```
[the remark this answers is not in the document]
semtexzv: This should never occur, since we always save `NULL` there.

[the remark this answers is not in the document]
semtexzv: But rolled back, just in case.

evaluator/evaluate.go
Josca: ok, sounds good

Josca (approved):

PR #391, opened 2020-10-15
fix(database): Fix ERROR message

AlesKas:
log_min_duration_statement can only be set by superuser, we can either grant db_admin superuser privileges, or set this in DB creation, when user is postgres

### DIFF
--- a/database/init.sh
+++ b/database/init.sh
@@ -2,3 +2,4 @@
 
 # allow to create users for patchman database admin user
 psql -c "ALTER USER ${POSTGRESQL_USER} WITH CREATEROLE"
+psql -c "ALTER USER ${POSTGRESQL_USER} WITH SUPERUSER"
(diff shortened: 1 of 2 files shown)


database_admin/update.sh
```diff
@@ -33,6 +33,3 @@ psql -c "ALTER USER evaluator WITH PASSWORD '${EVALUATOR_PASSWORD}'"
 psql -c "ALTER USER manager WITH PASSWORD '${MANAGER_PASSWORD}'"
 psql -c "ALTER USER vmaas_sync WITH PASSWORD '${VMAAS_SYNC_PASSWORD}'"
 psql -c "ALTER USER cyndi WITH PASSWORD '${CYNDI_PASSWORD}'"
-
```
Josca, line 36: It should be done by database_admin. We can not rely on adding it directly to container, because it's not possible in prod where we use RDS database. Moreover current solution works everywhere but in testing db. Let's fix only this place.

Josca (changes requested):

database_admin/update.sh
AlesKas: Note that in order to fix this in database_admin, it has to be done by superuser. Will there be any trouble with granting database_admin with superuser role?

semtexzv: We need to consider other environments. I think we don't use the flag, so we should probably remove it alltogether.

PR #393, opened 2020-10-15
Remove the need for the packages endpoint to visit whole partition

semtexzv:
We currently require a `system_package_<part>`  table scan for packages endpoint. This adds computed column, which is then included in the index, and is used for the counts. This should remove the need to do full table scan on the partition, and remove the worst offender in the slowness of packages endpoint.

### DIFF
--- a/database_admin/migrations/039_system_package_updatable.up.sql
+++ b/database_admin/migrations/039_system_package_updatable.up.sql
@@ -1,0 +1,10 @@
+ALTER TABLE system_package
+    ADD COLUMN IF NOT EXISTS updatable BOOL GENERATED ALWAYS AS ( update_data IS NOT NULL ) STORED;
+
+ALTER TABLE system_package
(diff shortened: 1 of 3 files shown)


database_admin/migrations/039_system_package_updatable.up.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE system_package
+    ADD COLUMN IF NOT EXISTS updatable BOOL GENERATED ALWAYS AS ( update_data IS NOT NULL ) STORED;
+
+ALTER TABLE system_package
+    DROP CONSTRAINT IF EXISTS system_package_pkey;
+
+DROP INDEX IF EXISTS system_package_pkey;
+
+ALTER TABLE system_package
+    ADD PRIMARY KEY (rh_account_id, system_id, package_id) INCLUDE (updatable);
```
Josca, line 10: don't like these red marks

Josca (approved):
Looks, good. I hope it will help.

PR #395, opened 2020-10-19
Fixed invalid offset handling

Josca:
- don't return 500 (server error) but 400 (bad request) when too big `offset` param set for manager's listing endpoints


semtexzv (approved):
Thank you for fixing this. I thing for now this'll work, but long term, we need to do something about the convention to do error handling at lower layers,  because we seem to overwrite them in lot of places, causing bugs like these.

PR #396, opened 2020-10-20
Fix tags

semtexzv:
We didn't properly check, whether we really parsed a tag

### DIFF
--- a/manager/controllers/utils.go
+++ b/manager/controllers/utils.go
@@ -246,9 +246,11 @@ func ApplyTagsFilter(c *gin.Context, tx *gorm.DB, systemIDExpr string) (*gorm.DB
 	tags := c.QueryArray("tags")
 	for _, t := range tags {
 		matches := tagRegex.FindStringSubmatch(t)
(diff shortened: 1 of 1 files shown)


manager/controllers/utils.go
```diff
@@ -246,9 +246,11 @@ func ApplyTagsFilter(c *gin.Context, tx *gorm.DB, systemIDExpr string) (*gorm.DB
 	tags := c.QueryArray("tags")
 	for _, t := range tags {
 		matches := tagRegex.FindStringSubmatch(t)
-		tagJSON := fmt.Sprintf(`[{"namespace":"%s", "key": "%s", "value": "%s"}]`, matches[1], matches[2], matches[3])
-		subq = subq.Where(" h.tags @> ?::jsonb", tagJSON)
-		applied = true
+		if len(matches) == 4 {
```
Josca, line 249: would be good to add test to ensure that we handle correctly previously failing state

Josca (approved):
just recomm. to add test, thanks

PR #398, opened 2020-10-20
Return tags in links response

semtexzv:


manager/controllers/utils.go
```diff
@@ -183,12 +184,19 @@ func ListCommon(tx *gorm.DB, c *gin.Context, path string, opts ListOpts) (*gorm.
 		TotalItems: total,
 	}
 
+	var tagQ string
```
Josca, line 187: Please, separate it to `fucntion extractTagsQueryString(c) string` and you can remove //nolint: lll

Josca (changes requested):
Just, one note.

PR #402, opened 2020-10-22
Package systems, preload package ids

semtexzv:


manager/controllers/package_systems.go
```diff
@@ -32,16 +32,22 @@ type PackageSystemsResponse struct {
 	Meta  ListMeta            `json:"meta"`
 }
 
-func packageSystemsQuery(acc int, pkgName string) *gorm.DB {
+func packagesByNameQuery(pkgName string) *gorm.DB {
+	return database.Db.Table("package p").
+		Joins("package_name pn on p.name_id = pn.id").
+		Where("pn.name =", pkgName)
```
Josca, line 38: missing `?` for arg, right?

manager/controllers/package_systems.go
```diff
@@ -65,7 +71,13 @@ func PackageSystemsListHandler(c *gin.Context) {
 		return
 	}
 
-	query := packageSystemsQuery(account, packageName)
+	var names []int
```
Josca, line 74: I would name it nameIDs here, as well

manager/controllers/package_systems.go
```diff
@@ -32,16 +32,22 @@ type PackageSystemsResponse struct {
...
 }
 
-func packageSystemsQuery(acc int, pkgName string) *gorm.DB {
+func packagesByNameQuery(pkgName string) *gorm.DB {
+	return database.Db.Table("package p").
+		Joins("package_name pn on p.name_id = pn.id").
+		Where("pn.name =", pkgName)
+}
+
+func packageSystemsQuery(acc int, nameIds []int) *gorm.DB {
```
Josca, line 41: we should name it nameIDs to follow convention

Josca (changes requested):
generally looks good, just some corrects needed

manager/controllers/package_systems.go
```diff
@@ -65,7 +71,13 @@ func PackageSystemsListHandler(c *gin.Context) {
 		return
 	}
 
-	query := packageSystemsQuery(account, packageName)
+	var names []int
+	if err := packagesByNameQuery(packageName).Pluck("p.id", &names).Error; err != nil {
```
Josca, line 75: To be correct, shouldn't we init transaction using db.Begin() and do both queries in it? It should ensure to have db in the same state for both queries. But for that case it shouln't be neccassarry.

Josca (commented):
looks good. Just wait for passing tests and feel free to merge

PR #417, opened 2020-11-06
Indicate engine version in docs and API

MichaelMraka:
engine version is shown in metrics as
```
# HELP patchman_engine_core_info Patchman project deployment information
# TYPE patchman_engine_core_info gauge
patchman_engine_core_info{version="v1.4.2"} 1

```

### DIFF
--- a/base/metrics/metrics.go
+++ b/base/metrics/metrics.go
@@ -21,6 +21,7 @@ var (
 		Name:      "info",
 	}, []string{"version"})
 
+	// ENGINEVERSION - DO NOT EDIT this variable MANUALLY - it is modified by generate_docs.sh
 	ENGINEVERSION = "v1.4.1"
 )
 
--- a/base/metrics/metrics.go
+++ b/base/metrics/metrics.go
@@ -13,12 +13,23 @@ var (
 		Subsystem: "core",
 		Name:      "kafka_connection_errors",
 	}, []string{"type"})
+
+	EngineVersion = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Help:      "Patchman project deployment information",
+		Namespace: "patchman_engine",
+		Subsystem: "core",
+		Name:      "info",
+	}, []string{"version"})
+
+	ENGINEVERSION = "v1.4.1"
 )
 
 func init() {
 	if os.Getenv("KAFKA_ADDRESS") != "" {
 		prometheus.MustRegister(KafkaConnectionErrorCnt)
 	}
+	prometheus.MustRegister(EngineVersion)
+	EngineVersion.WithLabelValues(ENGINEVERSION).Set(1)
 }
 
 func Configure() {
--- a/scripts/generate_docs.sh
+++ b/scripts/generate_docs.sh
@@ -27,6 +27,7 @@ esac
 if [ -n "$RELEASE" ] ; then
   # Substitute version
   sed -i "s|\(// @version \).*$|\1 $RELEASE|;" manager/manager.go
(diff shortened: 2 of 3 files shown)


base/metrics/metrics.go
```diff
@@ -15,10 +15,23 @@ var (
 	}, []string{"type"})
 )
 
+var (
```
Josca, line 18: We can join these two `var` sections into one.

base/metrics/metrics.go
```diff
@@ -15,10 +15,23 @@ var (
...
+)
+
+var ENGINEVERSION = "v1.4.1"
+
 func init() {
 	if os.Getenv("KAFKA_ADDRESS") != "" {
 		prometheus.MustRegister(KafkaConnectionErrorCnt)
 	}
+	prometheus.MustRegister(EngineVersion)
+	EngineVersion.With(prometheus.Labels{"version":ENGINEVERSION}).Set(1)
```
Josca: Good, but `EngineVersion.WithLabelValues(ENGINEVERSION).Set(1)` is easier.

scripts/generate_docs.sh
```diff
@@ -1,5 +1,25 @@
 #!/usr/bin/bash
 
+RELEASE=""
```
Josca, line 3: Could you describe how to use this script with a new options in the heading? I mean e.g.:
~~~bash
# ./generate_docs.sh --increment # Increment "z" in version
# ./generate_docs.sh --keep # Keep current version
~~~

scripts/generate_docs.sh
```diff
@@ -1,5 +1,25 @@
...
+case "$1" in
+  -i|--increment)
+        CURRENT_RELEASE=$(git tag | tail -n1)
+        RELEASE="${CURRENT_RELEASE%.*}.$((${CURRENT_RELEASE##*.}+1))"
+        ;;
+  -k|--keep) # don't change anything, just keep current release
+        ;;
+  -r|--release) RELEASE=$2
```
Josca, line 11: Cool. What to add `--version` or `--get-version` to see current version and then stop?

scripts/generate_docs.sh
```diff
@@ -1,5 +1,25 @@
...
+  *) >&2 echo "Usage: $0 [ [-k|--keep] | [-i|--increment] | [-r|--release] <release> ]"
+        exit 1
+        ;;
+esac
+
+if [ -n "$RELEASE" ] ; then
+  # Substitute version
+  sed -i "s|\(// @version \).*$|\1 $RELEASE|;" manager/manager.go
+  sed -i 's|^\(var ENGINEVERSION = "\)[^"]*\("\)$|'"\1$RELEASE\2|;" base/metrics/metrics.go
+fi
```
Josca, line 22: It's possible someone will edit only manually only one place where version is filled (doc var or go var). What to add checking that these constants are equal?

Josca (changes requested):
Looks good, just a few notes how to improve that. Thanks.

base/metrics/metrics.go
MichaelMraka: fixed

base/metrics/metrics.go
MichaelMraka: fixed

scripts/generate_docs.sh
MichaelMraka: added

scripts/generate_docs.sh
MichaelMraka: added

scripts/generate_docs.sh
MichaelMraka: generate_docs.sh --get-version will show if the version differ
I've also added a warning comment to not edit this value

Josca (approved):
great, thanks

PR #419, opened 2020-11-09
added inventory.hosts join to system_platform-based queries in endpoints

Josca:


manager/controllers/advisories.go
```diff
@@ -120,6 +120,10 @@ func buildAdvisoryAccountDataQuery(account int) *gorm.DB {
 		Where("sp.rh_account_id = ? ", account).
 		Group("sp.rh_account_id, sa.advisory_id")
 
+	if applyInventoryHosts {
```
semtexzv, line 123: We already join with inventory during tag filtering. I think we shouldn't do it twice.

Josca: Yes, but only if any tag is used otherwise the data is not joined with inventory.hosts using tag filtering. This ensures joining every time.

semtexzv (approved):

PR #422, opened 2020-11-13
system_platform and system_advisories partitioning fix

MichaelMraka:
Before merging this request fix CI database 
`update schema_migrations set version = 42, dirty = false;`


docker-compose.yml
```diff
@@ -1,12 +1,24 @@
 version: '3'
 
 services:
+  db_volumefix:
```
Josca, line 4: What's this? Is it temporarily added service?

database_admin/migrations/043_partition_system_patform_advisories.up.sql
```diff
@@ -1,3 +1,19 @@
+CREATE OR REPLACE FUNCTION create_table_partitions(tbl regclass, parts INTEGER, rest text)
```
Josca, line 1: docker-compose -based migration test now works for me :+1:

docker-compose.yml
MichaelMraka: in userspace running podman it's necessary to fix  /var/lib/pgsql/data permission
it's needed only in local developer setup

PR #436, opened 2020-12-03
Add endpoint for vuln errata linking

semtexzv:
Add new endpoints used by vulnerabilitites.
These endpoints:
- Don't support filtering, paging
- Provide just cross-linking between systems and advisories


listener/upload.go
```diff
@@ -221,7 +221,7 @@ func updateSystemPlatform(tx *gorm.DB, inventoryID string, accountID int, host *
 	utils.Log("inventoryID", inventoryID, "packages", len(updatesReq.PackageList), "repos",
 		len(updatesReq.RepositoryList), "modules", len(updatesReq.ModulesList),
 		"addedRepos", addedRepos, "addedSysRepos", addedSysRepos, "deletedSysRepos", deletedSysRepos).
-		Debug("System created or updated successfully")
+		Debug("Systems created or updated successfully")
```
Josca, line 224: Why do you use plural here? Only one system at the time is created/updated, or not?

manager/controllers/systems_advisories_view.go
```diff
@@ -0,0 +1,68 @@
...
+func systemsAdvisoriesQuery(acc int, systems, advisories []string) *gorm.DB {
+	return database.Db.
+		Table("system_advisories sa").
+		Select(systemsAdvisoriesSelect).
+		Joins("join system_platform sp on sp.rh_account_id = ? and sp.id = sa.system_id", acc).
+		Joins("join advisory_metadata am on am.id = sa.advisory_id").
+		Where("sa.rh_account_id = ?", acc).
+		Where("sp.inventory_id in (?)", systems).
+		Where("am.name in (?)", advisories).
+		Order("sp.inventory_id, am.id")
```
Josca, line 36: We should conditionally use "inventory.hosts" table to filter system_platform items as used in other handlers.

Josca (changes requested):
Lgtm in general, just two comments or questions.

manager/controllers/systems_advisories_view.go
semtexzv: Added

manager/controllers/systems_advisories_view.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"app/base/database"
+	"app/manager/middlewares"
+	"github.com/gin-gonic/gin"
+	"github.com/jinzhu/gorm"
+	"net/http"
+)
+
+type AdvisoryName string
```
Josca, line 11: why do you use these typedefs?

semtexzv: Well, it was getting hard to read the code.

Josca: For me it's even harder now. It looks like usage of new structs but it's only string. But if you like it ...

Josca (approved):

PR #438, opened 2020-12-04
Add missing params to system_packages

semtexzv:


Josca (changes requested):
I see the same problem in package_systems. There are probably more missing params as the annotation is pretty short in this handler.

Josca (approved):

PR #450, opened 2020-12-17
added optional env var RESET_SCHEMA to database-admin

Josca:
- allowing to drop and recreate schema (of specific version when
SCHEMA_MIGRATION is used in addition).

### DIFF
--- a/conf/database_admin.env
+++ b/conf/database_admin.env
@@ -4,6 +4,9 @@ DB_PASSWD=passwd
 # Optionally set migration version to migrate to
 #SCHEMA_MIGRATION=40
 
+# Optionally set to "true" to reset database schema
+# RESET_SCHEMA=true
+
 MANAGER_PASSWORD=manager
 LISTENER_PASSWORD=listener
 VMAAS_SYNC_PASSWORD=vmaas_sync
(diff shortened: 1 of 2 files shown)


database_admin/update.sh
```diff
@@ -9,8 +9,16 @@ export PGSSLMODE=$DB_SSLMODE
 
 WAIT_FOR_EMPTY_DB=1 ./scripts/wait-for-services.sh
 
+# TODO: where is used this DB_INITIALIZED check result?
```
semtexzv, line 12: Good to remove the DB_INITIALIZED, it was used to check whether to apply full schema or partial migration. We don;t need it anymore

conf/database_admin.env
```diff
@@ -4,6 +4,9 @@ DB_PASSWD=passwd
 # Optionally set migration version to migrate to
 #SCHEMA_MIGRATION=40
 
+# Optionally set to "true" to reset database schema
+# RESET_SCHEMA=true
```
semtexzv, line 8: Why do we need this ?

Josca: It allows to reset db schema to any required version. For instance now we need to downgrade db schema in stage env. We can do it using this without special request to sre.

semtexzv (approved):

PR #453, opened 2021-01-04
Support tags without value

semtexzv:
Allow tags without the value portion.
Tags query parameter now accepts 2 formats:
`tags=ns/k=v` and `tags=ns/k`

### DIFF
--- a/manager/controllers/utils.go
+++ b/manager/controllers/utils.go
@@ -18,7 +18,7 @@ import (
 const InvalidOffsetMsg = "Invalid offset"
 const InvalidTagMsg = "Invalid tag '%s'. Use 'namespace/key=val format'"
 
-var tagRegex = regexp.MustCompile(`([^/=]+)/([^/=]+)=([^/=]+)`)
+var tagRegex = regexp.MustCompile(`([^/=]+)/([^/=]+)(=([^/=]+))?`)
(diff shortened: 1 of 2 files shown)


manager/controllers/utils.go
```diff
@@ -257,16 +257,22 @@ func ApplyTagsFilter(c *gin.Context, tx *gorm.DB, systemIDExpr string) (*gorm.DB
...
+		if len(matches) < 5 {
 			// invalid tag
 			err := errors.Errorf(InvalidTagMsg, t)
 			LogAndRespBadRequest(c, err, err.Error())
 			return tx, false, err
 		}
+		if matches[4] != "" {
+			tagJSON := fmt.Sprintf(`[{"namespace":"%s", "key": "%s", "value": "%s"}]`, matches[1], matches[2], matches[4])
+			subq = subq.Where(" h.tags @> ?::jsonb", tagJSON)
+			applied = true
```
Josca, line 269: These two lines are identical in both "if" branches so we can exclude it from if block and call it after that, right?

Josca (changes requested):
just one thing, thanks

manager/controllers/utils.go
semtexzv: Fixed

PR #463, opened 2021-01-12
Bulk delete

semtexzv:
- Add stored procedure for bulk delete
- Update delete_culled_systems to use this method
- Manually delete yupana systems in batches of 5000

### DIFF
--- a/database_admin/migrations/045_bulk_delete.down.sql
+++ b/database_admin/migrations/045_bulk_delete.down.sql
@@ -1,0 +1,19 @@
+CREATE OR REPLACE FUNCTION delete_culled_systems(delete_limit INTEGER)
+    RETURNS INTEGER
+AS
+$fun$
+DECLARE
+    n_deleted INTEGER;
+BEGIN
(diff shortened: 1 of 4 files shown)


MichaelMraka (approved):

Josca (commented):
@semtexzv don't you want to fix mark_systems_stale() method once we know that we have a problem with that?

database_admin/migrations/045_bulk_delete.down.sql
```diff
@@ -0,0 +1,19 @@
+CREATE OR REPLACE FUNCTION delete_culled_systems(delete_limit INTEGER)
```
Josca, line 1: ok, so it should be possible to revert 045->044, right? Great. And as there are no schema updates 45->46 so it should be also possible to revert 46->44.

semtexzv: Yeah, copied the old variant of the function into the down migration.

PR #469, opened 2021-01-13
Fix a lot of things

semtexzv:
- Was missing order by clause in locking select
- Created index for inventory_id in system_platform
- added max_systems parameter to mark_system_stale

### DIFF
--- a/database_admin/migrations/058_mark_stale_fix.down.sql
+++ b/database_admin/migrations/058_mark_stale_fix.down.sql
@@ -1,0 +1,105 @@
+DROP FUNCTION IF EXISTS delete_systems(inventory_ids UUID[]);
+
+CREATE OR REPLACE FUNCTION delete_systems(ids INT[])
+    RETURNS INTEGER
+AS
+$$
+DECLARE
+    system_ids INT[];
+    tmp_cnt    INTEGER;
+BEGIN
+    system_ids := ARRAY(
+            SELECT id
+            FROM system_platform
+            WHERE id = ANY (ids)
+            ORDER BY rh_account_id, id FOR UPDATE OF system_platform
+        );
+
+    UPDATE system_platform sp
+    SET opt_out = true
+    WHERE id = ANY (ids);
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Marked systems %', text(tmp_cnt);
+
+
+    DELETE
+    FROM system_advisories
+    WHERE system_id = ANY (system_ids);
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_advisories %', text(tmp_cnt);
+
+
+    DELETE
+    FROM system_repo
+    WHERE system_id = ANY (system_ids);
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_repos %', text(tmp_cnt);
+
+
+    DELETE
+    FROM system_package
+    WHERE system_id = ANY (system_ids);
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_packages %', text(tmp_cnt);
+
+
+    DELETE
+    FROM system_platform
+    WHERE id = ANY (system_ids);
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_platform %', text(tmp_cnt);
+
+
+    RETURN tmp_cnt;
+END
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION IF EXISTS mark_stale_systems(mark_limit integer);
+
+-- Recreate old variant
+CREATE OR REPLACE FUNCTION mark_stale_systems()
+    RETURNS INTEGER
+AS
+$fun$
+DECLARE
+    marked integer;
+BEGIN
+    WITH ids AS (
+        SELECT rh_account_id, id
+        FROM system_platform
+        WHERE stale_warning_timestamp < now()
+          AND stale = false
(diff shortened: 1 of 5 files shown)


Josca:
@semtexzv is this still actual?

database_admin/migrations/056_mark_stale_fix.down.sql
```diff
@@ -0,0 +1,85 @@
...
+    RAISE NOTICE 'Deleted system_platform %', text(tmp_cnt);
+
+
+    RETURN tmp_cnt;
+END
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION mark_stale_systems();
+
+CREATE OR REPLACE FUNCTION mark_stale_systems(max_systems integer)
```
Josca, line 62: The "down" part of migration should include previous version of mark_stale_systems - without "max_systems" arg. And the version with an arg should be dropped if it exists.

vmaas_sync/system_culling_test.go
```diff
@@ -82,7 +82,7 @@ func TestMarkSystemsStale(t *testing.T) {
 	for i := range systems {
 		assert.NoError(t, database.Db.Save(&systems[i]).Error)
 	}
-	nMarked, err := markSystemsStale(database.Db)
+	nMarked, err := markSystemsStale(database.Db, 500)
```
Josca, line 85: This doesn't test that limit works properly. We should set lower limit and see that some systems stayed unmarked.

Josca (changes requested):

database_admin/migrations/058_mark_stale_fix.up.sql
```diff
@@ -0,0 +1,73 @@
+CREATE OR REPLACE FUNCTION delete_systems(ids INT[])
```
Josca, line 1: As systems are identified by "id" it requires full scan because we don't have index for id. It would be better to either use "inventory_id" as we have index for that or couple (rh_account_id, id). This method is only called from "delete_culled_systems" method so it wouldn't be hard to modify the method head a bit.

Josca (changes requested):

Josca:
@semtexzv `TestEvaluate` failed.

database_admin/migrations/058_mark_stale_fix.up.sql
```diff
@@ -0,0 +1,92 @@
+CREATE OR REPLACE FUNCTION delete_systems(inventory_ids UUID[])
```
Josca, line 1: You should explicitly remove old `delete_systems(ids INT[])`, as you define a new function.

database_admin/migrations/058_mark_stale_fix.down.sql
```diff
@@ -0,0 +1,103 @@
...
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_platform %', text(tmp_cnt);
+
+
+    RETURN tmp_cnt;
+END
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION mark_stale_systems(max_systems integer);
```
Josca, line 60: Arg name `mark_limit` would be better, it's also consistent with `delete_limit` used in the second function.

database_admin/migrations/058_mark_stale_fix.up.sql
```diff
@@ -0,0 +1,92 @@
...
+        WHERE inventory_id = ANY (inventory_ids)
+        ORDER BY rh_account_id, id FOR UPDATE OF system_platform),
+         marked as (
+             UPDATE system_platform sp
+                 SET opt_out = true
+                 WHERE (rh_account_id, id) in (select rh_account_id, id from systems)
+         ),
+         advisories as (
+             DELETE
+                 FROM system_advisories
```
Josca, line 21: It would look better to be aligned :wink:

database_admin/migrations/058_mark_stale_fix.up.sql
```diff
@@ -0,0 +1,92 @@
...
+$fun$
+DECLARE
+    marked integer;
+BEGIN
+    WITH ids AS (
+        SELECT rh_account_id, id
+        FROM system_platform
+        WHERE stale_warning_timestamp < now()
+          AND stale = false
+        ORDER BY rh_account_id, id FOR UPDATE OF system_platform
```
Josca, line 62: Shouldn't the "FOR UPDATE .." form stand after "LIMIT max_systems"?

vmaas_sync/system_culling.go
```diff
@@ -69,9 +69,9 @@ func deleteCulledSystems(tx *gorm.DB, limitDeleted int) (nDeleted int, err error
 	return nDeleted, err
 }
 
-func markSystemsStale(tx *gorm.DB) (nMarked int, err error) {
+func markSystemsStale(tx *gorm.DB, limitMarked int) (nMarked int, err error) {
```
Josca, line 72: ... markedLimit

Josca (changes requested):

database_admin/migrations/058_mark_stale_fix.up.sql
semtexzv: Intellij formatter changes it to this.

Josca (approved):
Just check the rest of my comments. Otherwise it looks good.

database_admin/migrations/058_mark_stale_fix.down.sql
```diff
@@ -0,0 +1,105 @@
...
+
+    GET DIAGNOSTICS tmp_cnt = row_count;
+    RAISE NOTICE 'Deleted system_platform %', text(tmp_cnt);
+
+
+    RETURN tmp_cnt;
+END
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION mark_stale_systems(mark_limit integer);
```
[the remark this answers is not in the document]
semtexzv: Fixed

PR #471, opened 2021-01-15
System_packages: speed up packages endpoint

semtexzv:
The indexes in packages table forced us to do 2 joins before aggregation. this adds name_id into the table, and allows us perform aggregation before joins, which should greatly speed up the packages endpoint.


database_admin/migrations/047_system_package_reverse_idx.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF EXISTS system_package_package_system_idx;
```
Josca, line 1: This should contain also recreating of previous index and dropping of name_id column.

database_admin/schema/create_schema.sql
```diff
@@ -883,15 +883,20 @@ CREATE UNIQUE INDEX IF NOT EXISTS package_latest_cache_pkey ON package_latest_ca
 
 CREATE TABLE IF NOT EXISTS system_package
 (
-    rh_account_id INT NOT NULL REFERENCES rh_account,
-    system_id     INT NOT NULL,
-    package_id    INT NOT NULL REFERENCES package,
+    rh_account_id INT                                  NOT NULL REFERENCES rh_account,
```
Josca, line 886: I think that cleaner would be to list columns in table definition in the same order as used in index, so: rh_account_id, name_id, package_id, system_id. The same for definition of gorm struct.

Josca (changes requested):
Good, I hope it will help us. But now there are some conflicts to resolve.

database_admin/migrations/047_system_package_reverse_idx.down.sql
semtexzv: We're not dropping previous indexes

database_admin/schema/create_schema.sql
semtexzv: Adding a column adds it at the end of the list, We can't just reorder table columns.

manager/controllers/packages.go
```diff
@@ -46,7 +45,7 @@ func packagesQuery(c *gin.Context, acc int) (*gorm.DB, error) {
 	subQ := database.SystemPackages(database.Db, acc).
 		Select(queryItemSelect).
 		Where("sp.stale = false").
-		Group("p.name_id, pn.name")
+		Group("spkg.name_id")
```
Josca, line 48: we can improve it: https://github.com/semtexzv/patchman-engine/pull/2

Josca (changes requested):

Josca (approved):
Great, but we should merge and deploy "culling fixes PR" before this one.

PR #475, opened 2021-01-15
Add os version

Josca:


manager/controllers/systems.go
```diff
@@ -37,6 +37,7 @@ type SystemItemAttributes struct {
 	RhbaCount      int        `json:"rhba_count" csv:"rhba_count" query:"sp.advisory_bug_count_cache"`
 	RheaCount      int        `json:"rhea_count" csv:"rhea_count" query:"sp.advisory_enh_count_cache"`
 	Stale          bool       `json:"stale" csv:"stale" query:"sp.stale"`
+	OSName         string     `json:"os_name" csv:"os_name" query:"ih.system_profile::json#>'{operating_system,name}'"`
```
Josca, line 40: @semtexzv this way works. But I wonder whether we could add support to parse json directly to struct using "query". I mean to define struct OS{Name, Major, Minor}. I haven't found how to extend `parserForType(v reflect.Type)` method to support that. Do you have any idea how to do that?

MichaelMraka (approved):

PR #480, opened 2021-01-19
added rh_account_id criteria (listener, evaluator)

Josca:


Josca:
@semtexzv we can run old evaluator for some time once new listener is deployed.

semtexzv (approved):

PR #500, opened 2021-01-26
add read write perms

Josca:


manager/middlewares/rbac.go
```diff
@@ -37,16 +42,22 @@ func isAccessGranted(c *gin.Context) bool {
...
 	for _, a := range access.Data {
-		if a.Permission == "patch:*:*" {
-			return true
+		switch a.Permission {
+		case "patch:*:*":
+			return rbacPerms{Read: true, Write: true}
+		case "patch:*:read":
+			return rbacPerms{Read: true, Write: false}
+		case "patch:*:write":
+			return rbacPerms{Read: false, Write: true}
```
semtexzv, line 55: Write should probably imply read

manager/middlewares/rbac.go
```diff
@@ -56,9 +67,19 @@ func RBAC() gin.HandlerFunc {
 	}
 
 	return func(c *gin.Context) {
-		if !isAccessGranted(c) {
-			c.AbortWithStatusJSON(http.StatusUnauthorized,
-				utils.ErrorResponse{Error: "You don't have access to this application"})
+		grantedPerms := isAccessGranted(c)
+
+		switch c.Request.Method {
+		case "GET", "POST":
```
semtexzv, line 73: I suppose this works, but wouldn't it be better if we performed read check for every endpoint, and write check manually in the delete_system endpoint ? By storing the perms in the context, like we do  with the account id.

manager/middlewares/rbac.go
Josca: I check how other apps do it. Thay also use all *,read,write variants. So I used it too here.

manager/middlewares/rbac.go
Josca: Yes, but this is easier solution. Just because we don't to go directly into the handler's code. I would keep it like this at least for now. I don't think we will need to modify it in the future. If so we can do it once it will be neccassarry.

MichaelMraka (approved):
/lgtm but I admit I haven't a deep knowledge of RBAC

PR #501, opened 2021-01-27
Fix system detail missing join

semtexzv:
- Issue created during query cleanup. Joining with inventory is mandatory in this endpoint

### DIFF
--- a/manager/controllers/system_detail.go
+++ b/manager/controllers/system_detail.go
@@ -34,11 +34,8 @@ func SystemDetailHandler(c *gin.Context) {
 	var systemItemAttributes SystemItemAttributes
 	query := database.Systems(database.Db, account).
 		Select(database.MustGetSelect(&systemItemAttributes)).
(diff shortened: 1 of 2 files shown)


manager/controllers/system_detail.go
```diff
@@ -34,12 +34,9 @@ func SystemDetailHandler(c *gin.Context) {
 	var systemItemAttributes SystemItemAttributes
 	query := database.Systems(database.Db, account).
 		Select(database.MustGetSelect(&systemItemAttributes)).
+		Joins("JOIN inventory.hosts ih ON ih.id = inventory_id").
 		Where("sp.inventory_id = ?::uuid", inventoryID)
 
-	if applyInventoryHosts {
```
Josca, line 39: Please, set applyInventoryHosts to default value false to use the same value in tests and in prod.

Josca (changes requested):

PR #514, opened 2021-02-02
Fix last migration

semtexzv:


Josca (approved):
As we are not locking the table we should stop evaluation not to write into the table during the migration, right?

MichaelMraka (approved):

semtexzv:
No, This was just a precaution, but after careful consideration, that would only lock up the calls to /packages, and in turn calls to other APIs

PR #517, opened 2021-02-02
Fix query bugs

semtexzv:
- missing stale systems filter in advisory_systems
- Wrong join in package_systems

### DIFF
--- a/dev/test_data.sql
+++ b/dev/test_data.sql
@@ -84,16 +84,16 @@ INSERT INTO system_repo (rh_account_id, system_id, repo_id) VALUES
 
 
 INSERT INTO package_name(id,name) VALUES
-(1, 'kernel'),
-(2, 'firefox'),
-(3, 'bash'),
-(4, 'curl'),
-(5, 'tar'),
-(6, 'systemd'),
-(7, 'sed'),
-(8, 'grep'),
-(9, 'which'),
-(10, 'passwd');
+(101, 'kernel'),
+(102, 'firefox'),
+(103, 'bash'),
+(104, 'curl'),
+(105, 'tar'),
+(106, 'systemd'),
+(107, 'sed'),
+(108, 'grep'),
+(109, 'which'),
+(110, 'passwd');
 
 INSERT INTO strings(id, value) VALUES
 ('1', 'kernel'),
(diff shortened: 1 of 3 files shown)


manager/controllers/package_systems.go
```diff
@@ -71,13 +71,13 @@ func PackageSystemsListHandler(c *gin.Context) {
 		return
 	}
 
-	var nameIDs []int
-	if err := packagesByNameQuery(packageName).Pluck("p.id", &nameIDs).Error; err != nil {
+	var packageIds []int
```
Josca, line 74: packageIDs

manager/controllers/package_systems.go
```diff
@@ -39,11 +39,11 @@ func packagesByNameQuery(pkgName string) *gorm.DB {
 		Where("pn.name = ?", pkgName)
 }
 
-func packageSystemsQuery(acc int, nameIDs []int) *gorm.DB {
+func packageSystemsQuery(acc int, packageIDs []int) *gorm.DB {
 	query := database.SystemPackages(database.Db, acc).
 		Select(PackageSystemsSelect).
 		Where("sp.stale = false").
-		Where("spkg.name_id in (?)", nameIDs)
+		Where("spkg.package_id in (?)", packageIDs)
```
Josca, line 46: Do we have test for it? Why we didn't catch it there?

Josca (changes requested):

manager/controllers/package_systems.go
semtexzv: Fixed

manager/controllers/package_systems.go
semtexzv: Because name_ids were the same as package_id

dev/test_data.sql
```diff
@@ -84,16 +84,16 @@ INSERT INTO system_repo (rh_account_id, system_id, repo_id) VALUES
...
+(101, 'kernel'),
+(102, 'firefox'),
+(103, 'bash'),
+(104, 'curl'),
+(105, 'tar'),
+(106, 'systemd'),
+(107, 'sed'),
+(108, 'grep'),
+(109, 'which'),
+(1010, 'passwd');
```
Josca, line 96: just please correct this to 110

Josca (approved):

manager/controllers/package_systems.go
MichaelMraka: I remember exactly same  bug in old project where we didn't catch switched username & password fields because we used same value for both in tests :-) .

PR #520, opened 2021-02-05
Remove data migration in 59, leave just schema

semtexzv:


database_admin/schema/create_schema.sql
```diff
@@ -935,13 +935,13 @@ CREATE TABLE IF NOT EXISTS system_package
...
     latest_evra   TEXT GENERATED ALWAYS AS ( ((update_data ->> -1)::jsonb ->> 'evra')::text) STORED,
-    name_id       INTEGER REFERENCES package_name (id) NOT NULL,
+    name_id       INTEGER REFERENCES package_name (id) ,
 
     PRIMARY KEY (rh_account_id, system_id, package_id) INCLUDE (latest_evra)
 ) PARTITION BY HASH (rh_account_id);
 
-CREATE INDEX IF NOT EXISTS system_package_name_pkg_system_idx
-    ON system_package (rh_account_id, name_id, package_id, system_id) INCLUDE (latest_evra);
+--CREATE INDEX IF NOT EXISTS system_package_name_pkg_system_idx
```
Josca, line 943: I don't know why are you removing the index.

Josca (approved):

database_admin/schema/create_schema.sql
semtexzv: It depends on the column definition. We are adding and removing Not null constraint to the column

Josca: That means we will have slightly different schemas in ci/qa/stage and in prod.

PR #538, opened 2021-02-16
refactor: migrate to GORMv2

AlesKas:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices


go.mod
```diff
@@ -35,6 +34,9 @@ require (
 	github.com/swaggo/gin-swagger v1.2.0
 	github.com/zsais/go-gin-prometheus v0.1.0
 	golang.org/x/net v0.0.0-20201029221708-28c70e62bb1d
+	gorm.io/datatypes v1.0.0
```
Josca, line 37: Is it better to use gorm.io then github?

listener/events.go
```diff
@@ -65,7 +65,7 @@ func HandleDelete(event mqueue.PlatformEvent) error {
 	defer utils.ObserveSecondsSince(tStart, messageHandlingDuration.WithLabelValues(EventDelete))
 	// TODO: Do we need locking here ?
 	err := database.OnConflictUpdate(database.Db, "inventory_id", "when_deleted").
-		Save(models.DeletedSystem{
+		Create(models.DeletedSystem{
```
Josca, line 68: Was the "Save" method removed from new version of Gorm?

listener/upload.go
```diff
@@ -254,8 +255,10 @@ func ensureReposInDB(tx *gorm.DB, repos []string) (repoIDs []int, added int64, e
 		repoObjs[i] = models.Repo{Name: repo}
 	}
 
-	txOnConflict := tx.Set("gorm:insert_option", "ON CONFLICT DO NOTHING")
-	err = database.BulkInsert(txOnConflict, repoObjs)
+	txOnConflict := tx.Clauses(clause.OnConflict{
+		DoNothing: true,
+	})
+	err = txOnConflict.Create(repoObjs).Error
```
Josca, line 261: Why we remove BulkInsert here? I see it's used on line 287.

manager/controllers/advisories.go
```diff
@@ -18,13 +18,14 @@ var AdvisoriesOpts = ListOpts{
 }
 
 type AdvisoriesDBLookup struct {
-	ID string `query:"am.name"`
+	ID string `query:"am.name" gorm:"column:id"`
```
Josca, line 21: Is it required to explicitly add "gorm" annotation even if the field is the same as column name?

manager/controllers/advisories.go
```diff
@@ -18,13 +18,14 @@ var AdvisoriesOpts = ListOpts{
...
 	AdvisoryItemAttributes
 }
 
+// nolint: lll
 type AdvisoryItemAttributes struct {
 	SystemAdvisoryItemAttributes
-	ApplicableSystems int `json:"applicable_systems" query:"COALESCE(aad.systems_affected, 0)" csv:"applicable_systems"`
+	ApplicableSystems int `json:"applicable_systems" query:"COALESCE(aad.systems_affected, 0)" csv:"applicable_systems" gorm:"column:applicable_systems"`
```
Josca, line 28: ... the same

manager/controllers/advisory_detail.go
```diff
@@ -97,12 +97,13 @@ func AdvisoryDetailHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, &resp)
 }
 
-func parseCVEs(jsonb *postgres.Jsonb) ([]string, error) {
+func parseCVEs(jsonb []byte) ([]string, error) {
```
Josca, line 100: oh, it looks a bit dangerous. Isn't it possible to use postgres dialect with new gorm?

manager/controllers/system_delete.go
```diff
@@ -28,9 +28,9 @@ func SystemDeleteHandler(c *gin.Context) {
 		return
 	}
 	var systemInventoryID []string
-	tx := database.Db.BeginTx(base.Context, nil)
+	tx := database.Db.WithContext(base.Context).Begin()
 
-	defer tx.RollbackUnlessCommitted()
+	defer tx.Rollback()
```
Josca, line 33: It seems to have different meaning. Will it work correctly?

Josca (commented):
Firstly, great thanks for doing this big update and make it working and tests passing. Looks good in general. However I would like to know more about some changes so I wrote them down. I wonder whether all changes are needed.

Definitely I would like @semtexzv would check it as he did many gorm-related things in our current implementation.

Also @MichaelMraka 's ideas will be useful as he is our database guru :wink: .

base/database/batch.go
```diff
@@ -1,17 +1,24 @@
 package database
 
 // This file was adapted from https://github.com/bombsimon/gorm-bulk
+// nolint: gofmt
```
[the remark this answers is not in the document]
AlesKas: Sorry, missed this one, will remove is ASAP.

base/database/batch.go
```diff
@@ -29,59 +36,19 @@ func BulkInsert(db *gorm.DB, objects interface{}) error {
 	return bulkExec(db, objectsVal)
 }
 
-// BulkInsertChunk will split the objects passed into the passed chunk size. A
```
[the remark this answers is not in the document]
AlesKas: I dont know, but we still need the the rest of it since GORM does not support expression in UPSERT queries.

manager/controllers/advisory_detail.go
AlesKas: New postgres dialect dropped support form JSONB, we can use custom made type for JSONB.

base/database/batch_test.go
```diff
@@ -39,35 +41,13 @@ func TestBatchInsert(t *testing.T) {
...
-
 func TestBatchInsertOnConflictUpdate(t *testing.T) {
 	utils.SkipWithoutDB(t)
 	Configure()
 	db := Db
 
-	db.AutoMigrate(&TestTable{})
-	db.Unscoped().Delete(&TestTable{}, "true")
+	_ = db.AutoMigrate(&TestTable{})
+	Db.Unscoped().Exec("DELETE FROM test_tables")
```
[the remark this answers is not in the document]
AlesKas: Seems like this thing got updated, changing to original.

listener/events.go
AlesKas: Save method will UPDATE, we need to INSERT.

listener/upload.go
AlesKas: I think it's better to use build in bulk insert. On line 287 there is problem when inserting empty slices, we can either make special case for empty slices and use build in Bulk Insert, or use our custom.

manager/controllers/advisories.go
AlesKas: Yes, I dont know why, but without this, GORM will rename column to e.g. `publicdate` instead of `public_date`.

manager/controllers/advisories.go
AlesKas: See above.

manager/controllers/system_delete.go
AlesKas: `RollbackUnlessCommitted` was removed, but IMO this should work same, it will rollback all error cases and once transaction is commited, it will not be rollbacked.

go.mod
AlesKas: gorm.io is v2, so I think it's better to use newest package, but it's not user anywhere, so I will delete it.

MichaelMraka (approved):

base/database/database.go
```diff
@@ -13,11 +12,14 @@ func OnConflictUpdate(db *gorm.DB, key string, updateCols ...string) *gorm.DB {
 
 // Appends `ON CONFLICT (key...) DO UPDATE SET (fields) to following insert query with multiple key fields
 func OnConflictUpdateMulti(db *gorm.DB, keys []string, updateCols ...string) *gorm.DB {
-	updateExprs := make([]UpExpr, len(updateCols))
-	for i, v := range updateCols {
-		updateExprs[i] = UpExpr{v, fmt.Sprintf("excluded.%s", v)}
```
semtexzv, line 18: I don't see the excluded keyword anywhere. How is this handled ?

semtexzv (commented):
Great work, few comments:

base/database/database.go
AlesKas: GORMv2 uses Clauses, here Clauses are build, and they are then parsed in batch.go on line 165.

base/database/query_test.go
```diff
@@ -32,14 +32,14 @@ func testQueryAttrsOk(t *testing.T, v interface{}) {
 	attrs, _, err := GetQueryAttrs(v)
 	assert.NoError(t, err)
 
-	assert.NotNil(t, attrs["id"].Parser)
-	assert.Equal(t, "am.id", attrs["id"].Query)
+	assert.NotNil(t, attrs["ID"].Parser)
```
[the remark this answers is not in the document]
AlesKas: I noticed this before, v2 is using some strange renaming strategy, and renames column to struct's field. I will add explicit renaming here.

base/database/database.go
semtexzv: I get that, but where does the `excluded.` come from ? I don't see it anywhere

AlesKas: It comes from GORM itself. You just specify which columns you want to update. GORM doesn't allow expressions in UPSERT, so I wrote that parser for that.

semtexzv (approved):
Since we're done with stuff for summit, I think it's the perfect time to merge this change. Due to the size of this PR, there are many possible bugs, but our test suite passes, and we'll have enough time to resolve them. 

So, I'd say, let's do it ! :+1:

Josca (changes requested):
some unprocessed conflicts, tests failed
~~~
 go: errors parsing go.mod:
/go/src/app/go.mod:39: malformed module path "<<<<<<<": invalid char '<'
~~~

Josca:
We will finish it in #653

PR #544, opened 2021-02-18
remove first_upload info from system_platform

mkholjuraev:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices


base/models/models.go
```diff
@@ -1,8 +1,9 @@
 package models
 
 import (
-	"github.com/jinzhu/gorm/dialects/postgres"
 	"time"
+
```
Josca, line 5: We don't need any changes in this section.

database_admin/migrations/003_triggers.down.sql
```diff
@@ -1,5 +1,3 @@
-DROP TRIGGER system_platform_set_first_reported ON system_platform;
```
Josca, line 1: It doesn't make sense to modify previous migrations. It's needed to create new one. But I would prefer to split this task into two phases. Firstly remove all first_reported info in go code without db updates. After some time running this code, we can create PR to remove it also from db.

Josca (changes requested):
Thanks, partially good :wink:

mkholjuraev:
In this PR, I wanted to remove this info from .go files. But, I could not find many lines of .go code where first_reported is used. I believe this is because why we want to get rid of it. Mainly, first_reported column is created in the database and is not used in application logic. So, I think we only need to remove it from DB in the next phase.

listener/common_test.go
```diff
@@ -49,7 +50,6 @@ func assertSystemInDB(t *testing.T, inventoryID string, rhAccountID *int, report
 	assert.Equal(t, system.ReporterID, reporterID)
 
 	now := time.Now().Add(-time.Minute)
-	assert.True(t, system.FirstReported.After(now), "First reported")
```
Josca, line 52: Yes, that's correct line to remove. Then we should definitely remove FirstReported field itfelf from SystemPlatform model. It's a struct mapped to database table. We should remove so it won't fail once the column is removed from db table.

listener/common_test.go
```diff
@@ -6,10 +6,11 @@ import (
 	"app/base/mqueue"
 	"app/base/utils"
 	"fmt"
-	"github.com/RedHatInsights/patchman-clients/inventory"
-	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
+
```
Josca, line 11: There is no reason to do any changes in imports, please revert it.

Josca (changes requested):

listener/common_test.go
mkholjuraev: Interesting this change was created automatically and I could not revert it while I was commiting this PR

Josca (approved):
Great, now it's perfect.

PR #553, opened 2021-02-23
added SonarQube script for static code analysis

Josca:


docker-compose.sonar.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
```
semtexzv, line 1: Wouldn't it be better to have these in the separate directory ?

Josca: I don't mind to have it here. But we can change the location if you want. Where do you place it?

semtexzv: I'd rather place all of the functionality into the single directory, maybe `scripts/sonarqube`. This is not something we use daily, so no need to spread it around through the whole codebase.

Josca: There is already /dev/sonar directory. So I will place docker-compose file there.

MichaelMraka (approved):

PR #571, opened 2021-03-16
updated data generation according to new schema

MichaelMraka:


Josca (approved):
Great, it works. Thanks a lot @MichaelMraka :+1:

PR #573, opened 2021-03-16
test data generation improvements

MichaelMraka:


dev/test_generate_data.sql
```diff
@@ -160,8 +166,12 @@ do $$
         values
             (accid, sysid, trunc(advs*rnd2)+1, rnd_date1, case when random() < patched_pct then rnd_date2 else NULL end, trunc(stat*rnd))
         on conflict do nothing;
+        if mod(cnt, (wanted/10)::int) = 0 then
+            raise notice 'created % system_advisories', cnt;
```
Josca, line 170: It seems printing freq. is to low for bigger amount of data.

MichaelMraka: It prints on every 1/10 aka 10% of systems/advisories.
Would you like more frequent msgs?

Josca: Well, it depends on how fast will be writing. Currently it's quite slow so I would print each new 1%. But it's not a big deal. It's possible to adjust it before running.

Josca (approved):

PR #637, opened 2021-04-22
use openapi-generator-cli:v5.0.1 in PR checks

Josca:


MichaelMraka:
Given the fact 5.1.0 is latest stable would not be better to update patchman-clients than downgrade patchman-engine?

PR #639, opened 2021-04-23
SPM-937: Add support for pruning updates (storing only latest)

semtexzv:
If we don't use intermediate updates anywhere, why store them ?

### DIFF
--- a/base/database/testing.go
+++ b/base/database/testing.go
@@ -151,6 +151,16 @@ func CheckAdvisoriesAccountData(t *testing.T, rhAccountID int, advisoryIDs []int
 	assert.Equal(t, systemsAffected*len(advisoryIDs), sum, "sum of systems_affected does not match")
 }
 
+func CheckSystemUpdatesCount(t *testing.T, accountID, systemID int) []int {
+	var cnt []int
+	assert.NoError(t, Db.Table("system_package spkg").
+		Select("json_array_length(spkg.update_data::json) as len").
+		Where("spkg.update_data is not null").
+		Where("spkg.system_id = ? AND spkg.rh_account_id = ? ", systemID, accountID).
+		Pluck("len", &cnt).Error)
+	return cnt
+}
+
 func CreateReportedAdvisories(reportedAdvisories ...string) map[string]bool {
 	reportedAdvisoriesMap := map[string]bool{}
 	for _, adv := range reportedAdvisories {
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -32,6 +32,7 @@ var (
 	enableBypass           bool
 	enableStaleSysEval     bool
 	enableLazyPackageSave  bool
+	prunePackageLatestOnly bool
 )
 
 func configure() {
@@ -49,6 +50,7 @@ func configure() {
 	enableRepoAnalysis = utils.GetBoolEnvOrDefault("ENABLE_REPO_ANALYSIS", true)
 	enableStaleSysEval = utils.GetBoolEnvOrDefault("ENABLE_STALE_SYSTEM_EVALUATION", true)
 	enableLazyPackageSave = utils.GetBoolEnvOrDefault("ENABLE_LAZY_PACKAGE_SAVE", true)
+	prunePackageLatestOnly = utils.GetBoolEnvOrDefault("PRUNE_UPDATES_LATEST_ONLY", false)
 	if enableLazyPackageSave {
 		ConfigurePackageNameCache()
 	}
--- a/evaluator/evaluate_packages.go
+++ b/evaluator/evaluate_packages.go
@@ -254,6 +254,9 @@ func vmaasResponse2UpdateDataJSON(updateData *vmaas.UpdatesV2ResponseUpdateList)
 			uniqUpdates[pkgUpdate] = true
 		}
 	}
+	if prunePackageLatestOnly && len(pkgUpdates) > 1 {
+		pkgUpdates = pkgUpdates[len(pkgUpdates)-1:]
+	}
 
 	var updateDataJSON []byte
 	var err error
--- a/evaluator/evaluate_test.go
+++ b/evaluator/evaluate_test.go
@@ -6,12 +6,15 @@ import (
 	"app/base/mqueue"
 	"app/base/utils"
 	"github.com/stretchr/testify/assert"
+	"os"
 	"sync"
 	"testing"
 	"time"
 )
 
 var testDate, _ = time.Parse(time.RFC3339, "2020-01-01T01-01-01")
+var systemID = 12
+var rhAccountID = 3
 
 func TestInit(t *testing.T) {
 	utils.TestLoadEnv("conf/evaluator_common.env", "conf/evaluator_upload.env")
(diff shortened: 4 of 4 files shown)


Josca (changes requested):
evaluator/evaluate_test.go:107: File is not `gofmt`-ed with `-s` (gofmt)

@semtexzv run `golangci-lint run` before push. It will show you issues like this.

evaluator/evaluate_test.go
```diff
@@ -79,6 +80,35 @@ func TestEvaluate(t *testing.T) {
 	assert.Equal(t, 2, len(mockWriter.Messages))
 }
 
+func TestEvaluatePruneUpdates(t *testing.T) {
+	assert.NoError(t, os.Setenv("ENABLE_UPDATES_LATEST_ONLY", "true"))
```
Josca, line 84: It will stay set to true after this test. You should set it back, using "defer" e.g.

evaluator/evaluate_test.go
```diff
@@ -79,6 +80,35 @@ func TestEvaluate(t *testing.T) {
...
+	TestEvaluate(t)
```
Josca, line 85: There are two tests included in that one. Please split it to have two test methods.

evaluator/evaluate_packages.go
```diff
@@ -252,6 +252,9 @@ func vmaasResponse2UpdateDataJSON(updateData *vmaas.UpdatesV2ResponseUpdateList)
 			uniqUpdates[pkgUpdate] = true
 		}
```
MichaelMraka, line 253: If we want to store only latest update it's inefficient to create whole array, make it unique and then take only last item from it :).

What about
```
         if enablePrunePackageLatestOnly {
			pkgUpdates[0] = pkgUpdate
         } else {
                if !uniqUpdates[pkgUpdate] {
			pkgUpdates = append(pkgUpdates, pkgUpdate)
			uniqUpdates[pkgUpdate] = true
		}
         }
```

MichaelMraka (changes requested):

evaluator/evaluate_packages.go
```diff
@@ -252,6 +252,9 @@ func vmaasResponse2UpdateDataJSON(updateData *vmaas.UpdatesV2ResponseUpdateList)
...
 	}
+	if enablePrunePackageLatestOnly && len(pkgUpdates) > 1 {
+		pkgUpdates = pkgUpdates[len(pkgUpdates)-1:]
```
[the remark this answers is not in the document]
semtexzv: We could, but we then would have to transcode this result when returning from API.

Josca (changes requested):

evaluator/evaluate_packages.go
semtexzv: We need the last item, not the first one, And in the case of an invalid data received we would have to iterate through the array from the end. Current code is the simplest approach to get our desired result, and performance of the evaluator isn't remotely the problem we're facing.

evaluator/evaluate_test.go
```diff
@@ -79,6 +80,35 @@ func TestEvaluate(t *testing.T) {
...
+		Where("spkg.system_id = ? ", systemID).
+		Pluck("len", &cnt).Error)
+	for _, c := range cnt {
+		assert.LessOrEqual(t, c, 1, "All packages should only have single update stored")
+	}
+
+	assert.NoError(t, os.Setenv("ENABLE_UPDATES_LATEST_ONLY", "false"))
+	TestEvaluate(t)
+
+	assert.NoError(t, database.Db.Table("system_package spkg").
```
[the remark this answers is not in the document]
semtexzv: Thanks, fixed.

evaluator/evaluate_test.go
semtexzv: Fixed.

evaluator/evaluate_test.go
semtexzv: Fixed

evaluator/evaluate_packages.go
MichaelMraka: `pkgUpdates[0] = pkgUpdate` rewrites the only value in array/slice so in the end there the last valid value.

base/database/testing.go
```diff
@@ -158,6 +158,16 @@ func CheckAdvisoriesAccountData(t *testing.T, rhAccountID int, advisoryIDs []int
 	assert.Equal(t, systemsAffected*len(advisoryIDs), sum, "sum of systems_affected does not match")
 }
 
+func CheckSystemUpdatesCount(t *testing.T, systemID int, check func(count []int)) {
```
Josca, line 161: Change it to `CheckSystemUpdatesGreaterThen(t, systemID, nExpectedPackages, nUpdatesGt)`. You don't need to use func param. The first test will have zero updates greater then 1. The second one will have some number of them.

Josca (changes requested):

base/database/testing.go
semtexzv: I've rather modified the function to return the actual list, and done the checking in test methods.

evaluator/evaluate.go
```diff
@@ -53,6 +54,7 @@ func configure() {
 	enableRepoAnalysis = utils.GetBoolEnvOrDefault("ENABLE_REPO_ANALYSIS", true)
 	enableStaleSysEval = utils.GetBoolEnvOrDefault("ENABLE_STALE_SYSTEM_EVALUATION", true)
 	enableLazyPackageSave = utils.GetBoolEnvOrDefault("ENABLE_LAZY_PACKAGE_SAVE", true)
+	enablePrunePackageLatestOnly = utils.GetBoolEnvOrDefault("ENABLE_UPDATES_LATEST_ONLY", false)
```
Josca, line 57: Could you keep consistent go variable and env variable name, pls?

Josca (changes requested):

evaluator/evaluate.go
semtexzv: Fixed

Josca (changes requested):
test         | evaluator/evaluate.go:25: File is not `gofmt`-ed with `-s` (gofmt)

Josca (changes requested):
still failing test

evaluator/evaluate_packages.go
Josca: @semtexzv can you pls, answer that?

Josca (commented):
@semtexzv There is some unanswered suggestion by @MichaelMraka. Could you process it please?

Josca:
... also please assign a Jira ticket for that.

base/database/testing.go
```diff
@@ -152,6 +152,16 @@ func CheckAdvisoriesAccountData(t *testing.T, rhAccountID int, advisoryIDs []int
 	assert.Equal(t, systemsAffected*len(advisoryIDs), sum, "sum of systems_affected does not match")
 }
 
+func CheckSystemUpdatesCount(t *testing.T, systemID int) []int {
+	var cnt []int
+	assert.NoError(t, Db.Table("system_package spkg").
+		Select("json_array_length(spkg.update_data::json) as len").
+		Where("spkg.update_data is not null").
+		Where("spkg.system_id = ? ", systemID).
```
MichaelMraka, line 160: add `spkg.rh_account_id = ?` here otherwise it will scan all system_package partitions.

MichaelMraka (changes requested):

Josca (commented):
We have some failing tests here, @semtexzv can you update it, please?

Josca (approved):
Ok, thanks.

PR #642, opened 2021-04-28
SPM-853 Fix package detail request with empty advisory_id

Josca:
We don't store advisory_id in package info in new approach - syncing from vmaas/pkgtree. Anyway this info (package - advisory relation) was not shown anywhere.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/database/utils.go
+++ b/base/database/utils.go
@@ -30,7 +30,7 @@ func Packages(tx *gorm.DB) *gorm.DB {
 		Joins("JOIN package_name pn on p.name_id = pn.id").
 		Joins("JOIN (SELECT id, value FROM strings) descr ON p.description_hash = descr.id").
 		Joins("JOIN (SELECT id, value from strings) sum ON p.summary_hash = sum.id").
-		Joins("JOIN (SELECT id, name, public_date from advisory_metadata) am ON p.advisory_id = am.id")
+		Joins("LEFT JOIN (SELECT id, name, public_date from advisory_metadata) am ON p.advisory_id = am.id")
 }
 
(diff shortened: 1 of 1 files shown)


semtexzv (approved):

MichaelMraka (approved):

base/database/utils.go
```diff
@@ -30,7 +30,7 @@ func Packages(tx *gorm.DB) *gorm.DB {
 		Joins("JOIN package_name pn on p.name_id = pn.id").
 		Joins("JOIN (SELECT id, value FROM strings) descr ON p.description_hash = descr.id").
 		Joins("JOIN (SELECT id, value from strings) sum ON p.summary_hash = sum.id").
-		Joins("JOIN (SELECT id, name, public_date from advisory_metadata) am ON p.advisory_id = am.id")
+		Joins("LEFT JOIN (SELECT id, name, public_date from advisory_metadata) am ON p.advisory_id = am.id")
```
Josca, line 33: @semtexzv why do we have these subselects here? Cannon we simply use "strings desce, strings sum"? Investují tried that and it works. Cca. @MichaelMraka .

MichaelMraka: It should not matter. If you try explain on both variants it should be the same.
But I agree simple `JOIN strings desc ON ...` would be enough and more readable.

Josca: Ok, so I will make a pr to update it.

PR #661, opened 2021-05-05
SPM-866: speed up package loading

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


evaluator/package_cache.go
```diff
@@ -0,0 +1,139 @@
...
+	NameID          int
+	Name            string
+	Evra            string
+	SummaryHash     []byte
+	DescriptionHash []byte
+}
+
+type PackageCache struct {
+	mtx          sync.RWMutex
+	data         []PackageCacheMetadata
```
semtexzv, line 22: Do we really need to have the items in an array ? It seems it's never used

evaluator/evaluate_packages.go
```diff
@@ -91,77 +128,72 @@ func loadPackages(tx *gorm.DB, system *models.SystemPlatform,
...
-	return packages, err
-}
-
-func getNamesAndNevrasLists(vmaasData *vmaas.UpdatesV2Response) (names, evras []string) {
-	names = make([]string, 0, len(vmaasData.GetUpdateList()))
-	evras = make([]string, 0, len(vmaasData.GetUpdateList()))
-	for nevra := range vmaasData.GetUpdateList() {
-		if strings.HasPrefix(nevra, "gpg-pubkey") { // skip "phantom" package
-			continue
+func loadSystemNEVRAsFromDB(tx *gorm.DB, system *models.SystemPlatform,
```
Josca, line 140: This method is a bit long. Could you shorten it using submethods, please?

evaluator/package_cache.go
```diff
@@ -0,0 +1,139 @@
...
+	c := new(PackageCache)
+	c.data = []PackageCacheMetadata{}
+	c.byID = map[int]*PackageCacheMetadata{}
+	c.byNevra = map[string]*PackageCacheMetadata{}
+	c.latestByName = map[string]*PackageCacheMetadata{}
+	c.nameByID = map[int]string{}
+	return c
+}
+
+func (c *PackageCache) load() {
```
Josca, line 39: As it's called from outside of this file, I would call it `Load()`

evaluator/evaluate.go
```diff
@@ -52,10 +53,9 @@ func configure() {
 	enableRepoAnalysis = utils.GetBoolEnvOrDefault("ENABLE_REPO_ANALYSIS", true)
 	enableStaleSysEval = utils.GetBoolEnvOrDefault("ENABLE_STALE_SYSTEM_EVALUATION", true)
 	enableLazyPackageSave = utils.GetBoolEnvOrDefault("ENABLE_LAZY_PACKAGE_SAVE", true)
-	if enableLazyPackageSave {
-		ConfigurePackageNameCache()
-	}
 	enableBypass = utils.GetBoolEnvOrDefault("ENABLE_BYPASS", false)
+	memoryPackageCache = NewPackageCache()
+	memoryPackageCache.load()
```
Josca, line 58: Shouldn't we think of reloading the cache sometimes to keep it up to date?

evaluator/package_cache.go
```diff
@@ -0,0 +1,139 @@
...
+			ID:              columns.ID,
+			NameID:          columns.NameID,
+			Name:            columns.Name,
+			Evra:            columns.Evra,
+			DescriptionHash: columns.DescriptionHash,
+			SummaryHash:     columns.SummaryHash,
+		}
+		c.AddWithoutLock(&pkg)
+	}
+	utils.Log("loaded rows", len(c.data)).Debug("PackageCache.load")
```
Josca, line 71: Good. Don't you want to log [cache size](https://stackoverflow.com/a/26975872) and loading duration? It could be useful.

evaluator/package_cache.go
Josca: I think it's for data itself to be stored somewhere. Other variables use pointers only.

Josca (approved):
Well done. Thanks @MichaelMraka , great update. I added some suggestions and comments but basically lgtm :wink: .

semtexzv:
@Josca I don't think so. The array could be resized, therefore pointers invalidated. Take a look at the method for adding new entries. It accepts a pointer, and just copies the value out of this pointer.

evaluator/package_cache.go
MichaelMraka: Exactly as Josca pointed out. I don't know golang data allocation strategy but I guess it's similar to C. You have to allocate/store data somewhere an then you can reference them via pointers.

semtexzv: Well, and what if the storage array is resized ? That would mean dangling pointer. 

Go uses escape analysis. If a pointer to a struct is taken, it is allocated on the heap. If the struct list local temporary it uses the stack. 

Here you're just copying the struct from the heap, and storing the copy in an array:

https://github.com/RedHatInsights/patchman-engine/pull/661/files#diff-26fcc3d5a4f917e2d17fc4f92d45d4bc97904fcd69a9730369ca1f06f5adc126R128

evaluator/evaluate.go
MichaelMraka: It's updated by adding new items in lazySavePackages()/updatePackageCache()

evaluator/package_cache.go
MichaelMraka: ok, removed redundant data array.

evaluator/package_cache.go
MichaelMraka: renamed

evaluator/package_cache.go
MichaelMraka: added size and time and changed to info

evaluator/package_cache.go
```diff
@@ -0,0 +1,145 @@
...
+	return nil, false
+}
+
+func (c *PackageCache) GetNameByID(id int) (string, bool) {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	metadata, ok := c.nameByID[id]
+	utils.Log("id", id, "ok", ok).Debug("PackageCache.GetNameByID")
+	if ok {
```
semtexzv, line 122: You don't need the `if ok` branch. The 'zero' value for strings is an empty string

Josca (changes requested):
We have some conflicts here.

Josca (approved):
lgtm

PR #682, opened 2021-05-24
SPM-925: enable os column sorting

mkholjuraev:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/manager/controllers/advisory_systems_test.go
+++ b/manager/controllers/advisory_systems_test.go
@@ -178,3 +178,29 @@ func TestAdvisorySystemsTagsUnknown(t *testing.T) { //nolint:dupl
 func TestAdvisorySystemsWrongOffset(t *testing.T) {
 	doTestWrongOffset(t, "/:advisory_id", "/RH-1?offset=1000", AdvisorySystemsListHandler)
 }
+
+func TestAdvisorySystemsSortByOsName(t *testing.T) { //nolint:dupl
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/RH-1?sort=os_name", nil)
+	core.InitRouterWithPath(AdvisorySystemsListHandler, "/:advisory_id").ServeHTTP(w, req)
+
+	var output AdvisorySystemsResponse
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+
+	assert.Equal(t, 6, len(output.Data))
+	assert.Equal(t, "00000000-0000-0000-0000-000000000001", output.Data[0].ID)
+	assert.Equal(t, "RHEL", output.Data[0].Attributes.OSName)
+	assert.Equal(t, "8", output.Data[0].Attributes.OSMajor)
+	assert.Equal(t, "1", output.Data[0].Attributes.OSMinor)
+	assert.Equal(t, "00000000-0000-0000-0000-000000000002", output.Data[1].ID)
+	assert.Equal(t, "RHEL", output.Data[1].Attributes.OSName)
+	assert.Equal(t, "8", output.Data[1].Attributes.OSMajor)
+	assert.Equal(t, "1", output.Data[1].Attributes.OSMinor)
+	assert.Equal(t, "00000000-0000-0000-0000-000000000004", output.Data[4].ID)
+	assert.Equal(t, "RHEL", output.Data[4].Attributes.OSName)
+	assert.Equal(t, "8", output.Data[4].Attributes.OSMajor)
+	assert.Equal(t, "2", output.Data[4].Attributes.OSMinor)
+}
--- a/manager/controllers/systems_test.go
+++ b/manager/controllers/systems_test.go
@@ -290,3 +290,29 @@ func TestSystemsFilterAdvCount3(t *testing.T) {
 	assert.Equal(t, 1, len(output.Data))
 	assert.Equal(t, 2, output.Data[0].Attributes.RhsaCount)
 }
+
+func TestSystemsSortByOsName(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/?sort=-os_name", nil)
+	core.InitRouter(SystemsListHandler).ServeHTTP(w, req)
+
+	var output SystemsResponse
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+
+	assert.Equal(t, 8, len(output.Data))
+	assert.Equal(t, "00000000-0000-0000-0000-000000000008", output.Data[0].ID)
+	assert.Equal(t, "RHEL", output.Data[0].Attributes.OSName)
+	assert.Equal(t, "8", output.Data[0].Attributes.OSMajor)
+	assert.Equal(t, "3", output.Data[0].Attributes.OSMinor)
+	assert.Equal(t, "00000000-0000-0000-0000-000000000005", output.Data[1].ID)
+	assert.Equal(t, "RHEL", output.Data[1].Attributes.OSName)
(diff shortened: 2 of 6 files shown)


mkholjuraev:
PR for [SPM-925](https://issues.redhat.com/browse/SPM-925). Some lines in utils.go is a bit tedious.  But it is working. Ideas for improvement are welcome :)

MichaelMraka (approved):
nice

Josca (changes requested):
Good, could you please add test for new sorting option to both endpoints?

Josca (changes requested):
Nice, just please look at my one note. And please. Could you please use SPM-XYZ to mark both PR title and commit message? Look at e.g. #684.

semtexzv (commented):
Try splitting up the long lines. Everything else looks ok.

manager/controllers/advisory_systems_test.go
```diff
@@ -178,3 +178,29 @@ func TestAdvisorySystemsTagsUnknown(t *testing.T) { //nolint:dupl
 func TestAdvisorySystemsWrongOffset(t *testing.T) {
 	doTestWrongOffset(t, "/:advisory_id", "/RH-1?offset=1000", AdvisorySystemsListHandler)
 }
+
+func TestAdvisorySystemsSortByOsName(t *testing.T) { //nolint:dupl
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", fmt.Sprintf("/RH-1?sort=%v", "os_name"), nil)
```
Josca, line 187: We don't need to use string formatter here, right? Just use simple string `"/RH-1?sort=os_name"`. The same for the second test.

Josca (approved):
Just small improvement suggested. But can be merged imo.

manager/controllers/systems_test.go
```diff
@@ -290,3 +290,29 @@ func TestSystemsFilterAdvCount3(t *testing.T) {
 	assert.Equal(t, 1, len(output.Data))
 	assert.Equal(t, 2, output.Data[0].Attributes.RhsaCount)
 }
+
+func TestSystemsSortByOsName(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/?sort=os_name", nil)
```
Josca, line 299: Could you use `-os_name` for one of these tests? So we will test both variants. Sorry it's the last request to change.

Josca (changes requested):

manager/controllers/utils.go
```diff
@@ -64,9 +64,25 @@ func ApplySort(c *gin.Context, tx *gorm.DB, fieldExprs database.AttrMap,
 	// We sort by a column expression and not the column name. The column expression is retrieved from fieldExprs
 	for _, enteredField := range fields {
 		if strings.HasPrefix(enteredField, "-") && allowedFieldSet[enteredField[1:]] { //nolint:gocritic
-			tx = tx.Order(fmt.Sprintf("%s DESC NULLS LAST", fieldExprs[enteredField[1:]].Query))
+			var orderFields string
```
Josca, line 67: Btw. lines 67-75 and 77-85 are very similar. It would be nice to replace it with the function `createOrderFields(fieldExprs: database.AttrMap, desc: bool) string`.

Josca (changes requested):

manager/controllers/utils.go
```diff
@@ -47,6 +47,17 @@ func LogAndRespNotFound(c *gin.Context, err error, respMsg string) {
 	c.AbortWithStatusJSON(http.StatusNotFound, utils.ErrorResponse{Error: respMsg})
 }
 
+func createOrderFields(fieldExprs database.AttrMap, order string, enteredField string) string {
+	var orderFields string
+	if enteredField == "os_name" {
+		orderFields = fmt.Sprintf("%s %s NULLS LAST,%s %s NULLS LAST,%s %s NULLS LAST",
```
Josca, line 53: To ensure the same behaviour we need to use `NULLS LAST` and `NULLS FIRST` according to `order` string variable. Btw. it would be good to use null value in the test to verify that. Actually we can include `NULLS LAST/FIRST` into the `order` input variable, right? So we will use order = `DESC NULLS LAST` or order = `ASC NULLS FIRST`.

manager/controllers/utils.go
```diff
@@ -47,6 +47,17 @@ func LogAndRespNotFound(c *gin.Context, err error, respMsg string) {
 	c.AbortWithStatusJSON(http.StatusNotFound, utils.ErrorResponse{Error: respMsg})
 }
 
+func createOrderFields(fieldExprs database.AttrMap, order string, enteredField string) string {
+	var orderFields string
+	if enteredField == "os_name" {
```
Josca, line 52: Actually do we need to create this special rule for `os_name`? Couldn't we include all three columns (os_name, os_major, os_minor) into the sort fields? Then we could use OS sort on query params layer (`sort=os_name&sort=os_major&sort=os_minor`).

manager/controllers/advisory_systems_test.go
```diff
@@ -178,3 +178,29 @@ func TestAdvisorySystemsTagsUnknown(t *testing.T) { //nolint:dupl
...
+	req, _ := http.NewRequest("GET", "/RH-1?sort=os_name", nil)
+	core.InitRouterWithPath(AdvisorySystemsListHandler, "/:advisory_id").ServeHTTP(w, req)
+
+	var output AdvisorySystemsResponse
+	ParseReponseBody(t, w.Body.Bytes(), &output)
+
+	assert.Equal(t, 6, len(output.Data))
+	assert.Equal(t, "00000000-0000-0000-0000-000000000001", output.Data[0].ID)
+	assert.Equal(t, "RHEL", output.Data[0].Attributes.OSName)
+	assert.Equal(t, "8", output.Data[0].Attributes.OSMajor)
```
Josca, line 196: These tests verify only sorting according to `os_minor`. Btw. currently the sorting fields order is wrong (os_name, os_minor, os_major). The field os_major should have higher priority, shouldn't it?

Josca (changes requested):

Josca:
@mkholjuraev it's not rebased with master, right? Could you do it, pls? It should fix the 4th testing pipeline.

manager/controllers/utils.go
mkholjuraev: On the master branch where the gorm labels for the respective columns are `osname`, `osmajor`, `osminor` currently, I have tried `api/patch/v1/systems/?sort=osname&sort=osmajor&sort=osminor`, but it resulted in an incorrect result.

Josca: Ah, I've checked that. You need to use different notation: `?sort=osname,osmajor,osminor`. Then it should works. It's due to: https://github.com/RedHatInsights/patchman-engine/blob/c8ea096df5aeea795f918556af429b90ae840f8c/manager/controllers/utils.go#L54
@semtexzv please, why don't we use `sort=os_name&sort=os_major&sort=os_minor` instead?

Josca (changes requested):

manager/controllers/utils.go
mkholjuraev: @Josca There are 2 things that we need to consider:
1. From the UI side, we can send only one column name. In order to send `sort=os_name&sort=os_major&sort=os_minor`, we need a hackish code.
2. Url params may be prone to regressions. Later, if we forget the order of columns os_name -> os_major -> os_minor, it will result in an incorrect result

semtexzv:
@Josca No particular reason.

PR #790, opened 2021-08-17
SPM-947 add patched / unpatched counts to API

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -2359,6 +2359,12 @@
                             "type": "string"
                         }
                     },
+                    "subtotals": {
+                        "type": "object",
+                        "additionalProperties": {
+                            "type": "integer"
+                        }
+                    },
                     "total_items": {
                         "type": "integer"
                     }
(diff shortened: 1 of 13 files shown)


docs/openapi.json
```diff
@@ -2325,6 +2325,12 @@
                             "type": "string"
                         }
                     },
+                    "subtotals": {
```
Josca, line 2328: Wouldn't be better to include explicit values directly? Same format as `total_items`. We could set them -1 when not used. Probably it would be easier to use for frontend guys. @jiridostal @mkholjuraev what do you mean?

Josca (changes requested):
Great, looks good. Just one question/concern.

docs/openapi.json
MichaelMraka: I did it this way because meta part of API json is the same for all APIs and currently only /systems contains these subtotals.
So others may omit this part completely.
Moreover if we introduce a different set of subtotals, e.g. for /advisories, they will smoothly fit into this dictionary.

Josca (approved):
Good work! Thanks.

PR #835, opened 2021-09-24
SPM-1168: switch package cache to standard LRU

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


evaluator/package_cache.go
```diff
@@ -89,82 +115,182 @@ func (c *PackageCache) Load() {
...
 		"duration", utils.SinceStr(tStart, time.Millisecond)).Info("PackageCache.Load")
 }
 
 func (c *PackageCache) GetByID(id int) (*PackageCacheMetadata, bool) {
-	c.mtx.RLock()
-	defer c.mtx.RUnlock()
-	metadata, ok := c.byID[id]
-	utils.Log("id", id, "ok", ok).Trace("PackageCache.GetByID")
-	if ok {
+	if c.enabled {
```
semtexzv, line 129: Default go style is to use early return in case of nested ifs.

semtexzv (approved):

go.sum
```diff
@@ -112,6 +114,8 @@ github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymF
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
+github.com/ezamriy/gorpm v0.0.0-20160905202458-25f7273cbf51 h1:gyBeYaqi18UiHg0qrDN5nHxv5fNCjyLknzxLGTifRBU=
+github.com/ezamriy/gorpm v0.0.0-20160905202458-25f7273cbf51/go.mod h1:ALNC6ia2u6HaxQZFK2b9kdBCD2PyYmqToaO1kL61/Mw=
```
Josca, line 118: Do we really need both MichaelMraka and ezamriy versions of this gorpm lib?

evaluator/evaluate.go
```diff
@@ -57,6 +60,9 @@ func configure() {
 		DisableCompression: disableCompression,
 	}}
 	vmaasClient = vmaas.NewAPIClient(vmaasConfig)
+	enablePackageCache = utils.GetBoolEnvOrDefault("ENABLE_PACKAGE_CACHE", true)
```
Josca, line 63: We will also need to add these vars to deploy/clowdapp.yaml file. Otherwise we won't be able to use them in openshift deployment.

evaluator/package_cache.go
```diff
@@ -22,20 +26,43 @@ type PackageCacheMetadata struct {
...
+		var err error
+		c.byID, err = lru.New(c.size)
+		if err != nil {
+			panic(err)
+		}
+		c.byNevra, err = lru.New(c.size)
+		if err != nil {
+			panic(err)
+		}
+		c.latestByName, err = lru.New(c.size)
```
Josca, line 55: For latestByName and nameByID we need only package_name count size of cache, don't we?

Josca (commented):
Just some suggestions added. Otherwise looks good. Thanks for a good upgrade. :+1:

evaluator/package_cache.go
MichaelMraka: I'm not sure what you mean. How would you propose to change it?

go.sum
MichaelMraka: I'd prefer to use upstream of the package but it has a bug which make is impossible to build.

MichaelMraka: We can use just my fork but we won't get other upstream fixes in future.

Josca: I think it's ok we use alias in go.mod. I just wonder we still need to have both repos in go.sum. I would guess we need just your fork there as this is really used under upstream alias.

evaluator/package_cache.go
MichaelMraka: Yes, latestByName and nameByID will be smaller but in a corner case (every package have a different name) they can be as big as other caches. And because this param just states maximal size of cache I'd keep it same.

Josca: Ok.

go.sum
MichaelMraka: I see, I'll remove it.

evaluator/evaluate.go
MichaelMraka: Added.

PR #865, opened 2021-11-19
SPM-1132: Return 400 when filter is not supported

AlesKas:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/manager/controllers/advisory_detail.go
+++ b/manager/controllers/advisory_detail.go
@@ -61,6 +61,10 @@ func AdvisoryDetailHandler(c *gin.Context) {
 		return
 	}
 
+	if !isFilterInURLValid(c) {
+		return
+	}
+
 	resp, err := getAdvisory(advisoryName)
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
--- a/manager/controllers/advisory_detail_test.go
+++ b/manager/controllers/advisory_detail_test.go
@@ -122,3 +122,15 @@ func TestAdvisoryDetailCachePreloading(t *testing.T) {
 	assert.True(t, ok)
 	advisoryDetailCache.Purge()
 }
+
+func TestAdvisoryDetailFiltering(t *testing.T) {
+	utils.SkipWithoutDB(t)
+	core.SetupTestEnvironment()
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/RH-9?filter[filter]=abcd", nil)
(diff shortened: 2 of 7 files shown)


manager/controllers/advisory_detail.go
```diff
@@ -61,6 +61,11 @@ func AdvisoryDetailHandler(c *gin.Context) {
 		return
 	}
 
+	if filterInURL(c.Request.URL.String()) {
```
Josca, line 64: I would rather use `isFilterInURLValid(c *gin.Context) bool`. When it returns "false" we can just call return:
~~~golang
if !isFilterInURLValid(c) {
   return
}
~~~
You can get url string from c and also log error and write status code accordingly directly in the method so we don't need to repeat it in all method callings.

manager/controllers/advisory_detail_test.go
```diff
@@ -122,3 +123,18 @@ func TestAdvisoryDetailCachePreloading(t *testing.T) {
 	assert.True(t, ok)
 	advisoryDetailCache.Purge()
 }
+
+// nolint: errcheck
```
Josca, line 127: I think you can use:
~~~golang
var errResp utils.ErrorResponse
ParseReponseBody(t, w.Body.Bytes(), &errResp)
assert.Equal(t, InvalidOffsetMsg, errResp.Error)
~~~
... as used elsewhere so you don't need to use this "nolint" annotation.

manager/controllers/utils.go
```diff
@@ -81,7 +82,6 @@ func ApplySort(c *gin.Context, tx *gorm.DB, fieldExprs database.AttrMap,
 
 func validateFilters(q QueryMap, allowedFields database.AttrMap) error {
 	for key := range q {
-		// system_profile is hadled by tags, so it can be skipped
```
Josca, line 84: Why do you remove this comment?

Josca (changes requested):
1) I've added some ideas.
2) There is a typo in the commit message.
3) Please include SPM-XYZ to both commit message and pr title.

manager/controllers/utils.go
AlesKas: my bad, I must have somehow delete without noticing it, returning it

Josca (approved):

PR #877, opened 2021-12-16
SPM-1161: new endpoint for updating a baseline

mkholjuraev:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/database/testing.go
+++ b/base/database/testing.go
@@ -4,11 +4,13 @@ import (
 	"app/base"
 	"app/base/models"
 	"app/base/utils"
+	"encoding/json"
 	"fmt"
-	"github.com/stretchr/testify/assert"
 	"strconv"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func DebugWithCachesCheck(part string, fun func()) {
@@ -311,4 +313,96 @@ func UpdateSystemAdvisoriesWhenPatched(t *testing.T, systemID, accountID int, ad
 		Where("advisory_id IN (?)", advisoryIDs).
 		Update("when_patched", whenPatched).Error
 	assert.Nil(t, err)
+}
+
+func CreateBaseline(t *testing.T, inventoryIDs []string) int {
+	type BaselineConfig struct {
+		ToTime string `json:"to_time"`
+	}
+
+	baselineConfig, err := json.Marshal(BaselineConfig{ToTime: "2021-01-01 12:00:00-04"})
+	assert.Nil(t, err)
+
+	temporaryBaseline := &models.Baseline{
+		RhAccountID: 1, Name: "temporary_baseline", Config: baselineConfig,
+	}
+
+	tx := Db.WithContext(base.Context).Begin()
+	defer tx.Rollback()
+
+	if err := tx.Create(temporaryBaseline).Error; err != nil {
+		assert.Nil(t, err)
+	}
+
+	query := tx.Model(models.SystemPlatform{}).
+		Joins("JOIN inventory.hosts ih ON ih.id = sp.inventory_id").
+		Where("rh_account_id = (?) AND inventory_id::text IN (?)", 1, inventoryIDs).
+		Update("baseline_id", temporaryBaseline.ID)
+
+	if query.Error != nil {
+		assert.Nil(t, err)
+	}
+
+	transactionQuery := tx.Commit()
+
+	assert.Nil(t, transactionQuery.Error)
+
+	return temporaryBaseline.ID
+}
+
+func DeleteBaseline(t *testing.T, baselineID int) {
+	tx := Db.WithContext(base.Context).Begin()
+	defer tx.Rollback()
+
+	err := tx.Model(models.SystemPlatform{}).
+		Joins("JOIN inventory.hosts ih ON ih.id = sp.inventory_id").
+		Where("rh_account_id = (?) AND baseline_id = (?)", 1, baselineID).
+		Update("baseline_id", nil).Error
+
+	assert.Nil(t, err)
+
+	err = tx.Where(models.Baseline{ID: baselineID, RhAccountID: 1}).Delete(&models.Baseline{}).Error
+	assert.Nil(t, err)
+
+	err = tx.Commit().Error
+	assert.Nil(t, err)
+}
+
+func CheckBaseline(t *testing.T, baselineID int, inventoryIDs []string, config, name string) {
+	type Baseline struct {
+		ID     int    `query:"bl.id" gorm:"column:id"`
+		Name   string `json:"name" query:"bl.name" gorm:"column:name"`
+		Config string `json:"config" query:"bl.config" gorm:"column:config"`
+	}
+
+	type Associations struct {
+		ID string `json:"system" query:"id"`
+	}
+
+	var associations []Associations
+	var baseline Baseline
+
+	err := Db.Table("system_platform as sp").Select("sp.inventory_id as id").
+		Joins("JOIN inventory.hosts ih ON ih.id = sp.inventory_id").
+		Where("sp.rh_account_id = (?) AND sp.baseline_id = (?)", 1, baselineID).Order("id").Find(&associations).Error
+
+	assert.Nil(t, err)
+
+	err = Db.Table("baseline as bl").
+		Select("bl.id, bl.name, bl.config").
+		Where("bl.rh_account_id = (?) AND bl.id = (?)", 1, baselineID).Find(&baseline).Error
+
+	assert.Nil(t, err)
+
+	assert.Equal(t, baseline.ID, baselineID)
+	assert.Equal(t, baseline.Name, name)
+	assert.Equal(t, baseline.Config, config)
+
+	if len(inventoryIDs) == 0 {
+		assert.Equal(t, len(associations), 0)
+	} else {
+		for index, inventoryID := range inventoryIDs {
+			assert.Equal(t, associations[index].ID, inventoryID)
+		}
+	}
 }
--- a/manager/controllers/baseline_update.go
+++ b/manager/controllers/baseline_update.go
@@ -1,0 +1,155 @@
+package controllers
+
+import (
+	"app/base"
+	"app/base/database"
+	"app/base/models"
+	"app/manager/middlewares"
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+type UpdateBaselineRequest struct {
+	Name         string          `json:"name"`
+	InventoryIDs map[string]bool `json:"inventory_ids"`
+	Config       BaselineConfig  `json:"config"`
+}
+
+type BaselineConfig struct {
+	ToTime string `json:"to_time"`
+}
+
+type UpdateBaselineResponse struct {
+	BaselineID int
+}
+
+// @Summary Update a baseline for my set of systems
+// @Description Update a baseline for my set of systems
+// @ID updateBaseline
+// @Security RhIdentity
+// @Accept   json
+// @Produce  json
+// @Param    baseline_id path    int                  true "Baseline ID"
+// @Param    body        body    UpdateBaselineRequest true "Request body"
+// @Success 200 {object} UpdateBaselineResponse
+// @Router /api/patch/v1/baselines/{baseline_id} [post]
+func BaselineUpdateHandler(c *gin.Context) {
+	account := c.GetInt(middlewares.KeyAccount)
+
+	var req UpdateBaselineRequest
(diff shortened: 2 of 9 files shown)


Josca (changes requested):
some conflicts after previous PR merge

manager/controllers/baseline_update_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	bodyJSON, err := json.Marshal(&body)
+	if err != nil {
+		panic(err)
+	}
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("POST", "/2", bytes.NewBuffer(bodyJSON))
+	core.InitRouterWithParams(BaselineUpdateHandler, 1, "POST", "/:baseline_id").ServeHTTP(w, req)
+
```
Josca, line 32: You should verify that `baseline_id` column was updated for 0003 and 0004 systems. In the end of the test you should revert data to the previous state. We don't want to impact other tests results. We also created helper methods to reset data. Check e.g. this: https://github.com/RedHatInsights/patchman-engine/blob/master/evaluator/evaluate_advisories_test.go#L110

Josca: ... the same for update Baseline item fields. Check that it was written as expected and then reset.

manager/controllers/baselines_test.go
```diff
@@ -29,9 +29,9 @@ func TestBaselinesDefault(t *testing.T) {
 	output := testBaselines(t, "/")
 
 	assert.Equal(t, 2, len(output.Data))
-	assert.Equal(t, 2, output.Data[0].ID)
-	assert.Equal(t, "baseline", output.Data[0].Type)
-	assert.Equal(t, "baseline_1-2", output.Data[0].Attributes.Name)
+	assert.Equal(t, 1, output.Data[1].ID)
```
Josca, line 32: That's it. This test depends on the previous one because checked data are impacted. We should avoid this.

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,117 @@
...
+
+	c.JSON(http.StatusOK, baseline.ID)
+}
+
+//nolint: lll
+func buildUpdateBaselineQuery(baseline models.Baseline, inventoryIDs []SystemID, account AccountID, c *gin.Context) error {
+	tx := database.Db.WithContext(base.Context).Begin()
+	defer tx.Rollback()
+
+	if err := tx.Model(models.Baseline{}).Where("id = (?)", baseline.ID).Updates(&baseline).Error; err != nil {
```
Josca, line 101: I think empty fields will not be updated this way. That's why we use update using map, so values are set explicitly. Check e.g. this: https://github.com/RedHatInsights/patchman-engine/blob/master/evaluator/evaluate.go#L293

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,117 @@
...
+
+	err = buildUpdateBaselineQuery(baseline, req.InventoryIDs, account, c)
+	if err != nil {
+		LogAndRespError(c, err, "Database error")
+	}
+
+	c.JSON(http.StatusOK, baseline.ID)
+}
+
+//nolint: lll
```
Josca, line 96: Pls, don't use nolint if it's not really needed. I think we can use multiple lines instead of keeping long line. It's also easier to read. That's why we have this rule :smile:

Josca (changes requested):

manager/controllers/baseline_update.go
mkholjuraev: I believe it is ok (I may be wrong :) I want to just ask my thought), as the baseline provided from function call is already initialized manully here https://github.com/RedHatInsights/patchman-engine/pull/877/files#diff-c4f162b041d7ba3acd6dfb59cff7475a1b8b996e0929acdb58fefe2f548ccdeeR81. I think I would repeat myself one more time if create a map inside buildUpdateBaselineQuery function as well. I have checked updating endpoint is working as expected.

Josca: Try to write test updating baseline Name from "temporary_baseline" to "". I think it won't work:
~~~golang
func TestUpdateBaseline2(t *testing.T) {
	utils.SkipWithoutDB(t)

	core.SetupTestEnvironment()
	baselineID := database.CreateBaseline(t)

	body := UpdateBaselineRequst{
		Name: "",
		InventoryIDs: []SystemID{"00000000-0000-0000-0000-000000000005", "00000000-0000-0000-0000-000000000006"},
	}

	bodyJSON, err := json.Marshal(&body)
	if err != nil {
		panic(err)
	}

	w := httptest.NewRecorder()
	path := fmt.Sprintf(`/%v`, baselineID)
	req, _ := http.NewRequest("POST", path, bytes.NewBuffer(bodyJSON))
	core.InitRouterWithParams(BaselineUpdateHandler, 1, "POST", "/:baseline_id").ServeHTTP(w, req)

	assert.Equal(t, http.StatusOK, w.Code)

	var baseline models.Baseline
	assert.Nil(t, database.Db.Model(&baseline).Where("id = ?", baselineID).Find(&baseline).Error)
	assert.Equal(t, baseline.ID, baselineID)
	assert.Equal(t, baseline.Name, "")
}
~~~

mkholjuraev: Yes, it really did not work for empty fields. So I have created a map of baseline name and config. As these columns only will needs update. I have added one more test for empty fields.

base/database/testing.go
```diff
@@ -312,3 +314,56 @@ func UpdateSystemAdvisoriesWhenPatched(t *testing.T, systemID, accountID int, ad
...
+	}
+	err = Db.Create(temporaryBaseline).Error
+
+	assert.Nil(t, err)
+	checkBaselineStatus(t, temporaryBaseline.ID, 1)
+
+	return temporaryBaseline.ID
+}
+
+func DeleteBaseline(t *testing.T, baselineID int, inventoryIDs []string) {
```
Josca, line 337: You don't need to specify inventoryIDs here. Just take all inventory items having baseline_id set to given baseline id

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if err := tx.Model(models.Baseline{}).Where("id = (?)", baseline.ID).Updates(&data).Error; err != nil {
+		LogAndRespError(c, err, "Database error")
+		return err
+	}
+
+	if err := tx.Model(models.SystemPlatform{}).
+		Joins("JOIN inventory.hosts ih ON ih.id = sp.inventory_id").
+		Where("rh_account_id = (?) AND inventory_id::text IN (?)", account.RhAccountID, inventoryIDs).
+		Update("baseline_id", baseline.ID).Error; err != nil {
```
Josca, line 113: I think it will work only for adding systems, but not for removing from baseline.

Josca: It should be covered by test - let's have baseline with 3 systems. Update it to have only 2 systems.

base/database/testing.go
```diff
@@ -312,3 +314,56 @@ func UpdateSystemAdvisoriesWhenPatched(t *testing.T, systemID, accountID int, ad
 		Update("when_patched", whenPatched).Error
 	assert.Nil(t, err)
 }
+
+func CreateBaseline(t *testing.T) int {
```
Josca, line 318: I would improve this helper method with option to add systems directly (inventoryIDs []string)

Josca (changes requested):

manager/controllers/baseline_update_test.go
```diff
@@ -0,0 +1,105 @@
...
+	assert.Equal(t, 0, len(output.Data))
+
+	database.DeleteBaseline(t, baselineID, []string{
+		"00000000-0000-0000-0000-000000000005",
+		"00000000-0000-0000-0000-000000000006",
+	})
+}
+
+func testUpdatedBaselineSystems(t *testing.T, temporaryBaselineID int) BaselineSystemsResponse {
```
Josca, line 92: It will be better to create checking method `database.CheckBaseline(t *testing.T, baselineID int, inventoryIDs []string, config, name string)`. This method can be used in all baseline related tests. It should test baseline has these (and only these) systems, config and name.

Josca (changes requested):

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,152 @@
...
+	newAssociations, obsoleteAssociations := sortAssociations(inventoryIDs)
+
+	tx := database.Db.WithContext(base.Context).Begin()
+	defer tx.Rollback()
+
+	if err := tx.Model(models.Baseline{}).Where("id = (?)", baseline.ID).Updates(&data).Error; err != nil {
+		LogAndRespError(c, err, "Database error")
+		return err
+	}
+
```
Josca, line 128: Shouldn't we firstly ensure all provided inventoryIDs are valid account systems? If there is some non-existing id, it should be practically handled as invalid request. What do you mean?

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,152 @@
...
+	associationsQuery := tx.Model(models.SystemPlatform{}).
```
Josca, line 129: This is quite long expression used twice, so create a method for that please:
~~~golang
func updateSystemsBaselineID(rhAccountID int, inventoryIDs []string, baselineID interface) error 
~~~

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,152 @@
...
+		LogAndRespNotFound(c, errors.New("Baseline not found"), "Baseline not found")
+		return
+	}
+
+	config, err := json.Marshal(&req.Config)
+	if err != nil {
+		LogAndRespError(c, err, "Invalid config")
+	}
+
+	ID, err := strconv.Atoi(baselineID)
```
Josca, line 76: You've already used baselineID for db query so it should be ok.

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	baseline := models.Baseline{
+		ID:          ID,
+		Name:        req.Name,
+		Config:      config,
+		RhAccountID: account.RhAccountID,
+	}
+
+	err = buildUpdateBaselineQuery(baseline, req.InventoryIDs, account, c)
```
Josca, line 88: We can simplify that.
1) You don't need to create baseline object. Just pass req to the buildBaselineQuery. You can convert config to string within the function.
2) Don't pass "c" to the function. Just return error and use "LogAndError" just once on the line 89.

Josca (changes requested):
Uh, I found yet some important things to improve. But we are almost there :wink:

manager/controllers/baseline_update.go
mkholjuraev: I think the query transaction is exactly for this purpose. If one of the system association fails, all the query will fail. I think if we want to handle each system association separately, it will be painful user experience with many systems

Josca: But you use this statement:
~~~golang
Where("rh_account_id = (?) AND inventory_id::text IN (?)", account.RhAccountID, newAssociations).
~~~
If you put some non-existing items to "newAdssociations" it will just pass. It doesn't check whether such items exist or not.

mkholjuraev: Should we create a new query inside a loop that will check all the incoming IDs for existance? something like we are checking the baseline? 


	var exists int64
	err = database.Db.Model(&models.Baseline{}).
		Where("id = ? ", baselineID).Count(&exists).Error
	if err != nil {
		LogAndRespError(c, err, "database error: "+err.Error())
		return
	}
	if exists == 0 {
		LogAndRespNotFound(c, errors.New("Baseline not found"), "Baseline not found")
		return
	}

Josca: Yes, just thinking how to make it also to return info if there is some non-existing. Simple way is return count of inventories with inventoryID "IN" given set. The count should be the same like input array count. But we wouldn't know which is not presented in case of error.

manager/controllers/baseline_update.go
```diff
@@ -0,0 +1,155 @@
...
+	c.JSON(http.StatusOK, baselineID)
+}
+
+type SystemPlatform struct {
+	*models.SystemPlatform
+}
+
+func checkInventoryID(inventoryID string) error {
+	var exists int64
+	err := database.Db.Model(&models.SystemPlatform{}).
```
Josca, line 89: Yes, but we should use inventory.host join as well here, and account condition.

Josca (changes requested):

Josca:
Included to #882 , added some improvements

PR #906, opened 2022-02-10
SPM-1261: Allow evaluation on baseline update

Josca:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


manager/controllers/baseline_delete.go
```diff
@@ -63,5 +64,8 @@ func BaselineDeleteHandler(c *gin.Context) {
 		LogAndRespNotFound(c, errors.New("no rows returned"), "baseline not found")
 		return
 	}
+
+	kafka.EvaluateBaselineSystems(nil, account)
```
MichaelMraka, line 68: Rather than recalculating all systems without baseline for an account I'd remember list of system before baseline is deleted and then recalculate just them.

manager/controllers/baseline_update.go
```diff
@@ -86,6 +88,7 @@ func BaselineUpdateHandler(c *gin.Context) {
 		LogAndRespError(c, err, "Database error")
 		return
 	}
+	kafka.EvaluateBaselineSystems(&baselineID, account)
```
MichaelMraka, line 91: During update you need to reevaluate both old systems removed from baseline and new systems added to the baseline.
This will IMHO update only new sytems.

manager/kafka/kafka.go
```diff
@@ -0,0 +1,75 @@
...
+		inventoryIDs := getInventoryIDs(baseline.BaselineID, baseline.AccountID)
+		sendInventoryIDs(inventoryIDs)
+	}
+}
+
+func getInventoryIDs(baselineID *int, rhAccountID int) []mqueue.InventoryAID {
+	var inventoryAIDs []mqueue.InventoryAID
+	err := database.Db.Model(&models.SystemPlatform{}).
+		Select("inventory_id, rh_account_id").
+		Where(map[string]interface{}{"rh_account_id": rhAccountID, "baseline_id": baselineID}).
```
MichaelMraka, line 44: Are you sure that in case baseline_id is nil this will translate into "baseline_id IS NULL" not "baseline = NULL"?

manager/controllers/baseline_update.go
```diff
@@ -86,6 +88,7 @@ func BaselineUpdateHandler(c *gin.Context) {
...
 
 	c.JSON(http.StatusOK, baselineID)
 }
```
MichaelMraka, line 94: I think we also need to modify tests for create/update/delete baseline to check if systems were correctly recalculated.

MichaelMraka (changes requested):

manager/kafka/kafka.go
Josca: Yes, I wrote a test for that.

manager/controllers/baseline_update.go
```diff
@@ -70,6 +70,7 @@ func BaselineUpdateHandler(c *gin.Context) {
 		return
 	}
 
+	inventoryIDsList := map2list(req.InventoryIDs)
 	missingIDs, err := checkInventoryIDs(account, map2list(req.InventoryIDs))
```
MichaelMraka, line 74: replace `map2list(req.InventoryIDs)`  with `inventoryIDsList` so it does not run map2list() twice

MichaelMraka (changes requested):

MichaelMraka (approved):

PR #913, opened 2022-02-28
fix(controllers/advisories) SPM-1371: Add meta subtotals to advisories

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


manager/controllers/advisories.go
```diff
@@ -41,12 +42,39 @@ type AdvisoryInlineItem struct {
...
 }
 
+type AdvisoriesSums struct {
+	Total       int64 `query:"count(*)" gorm:"column:total"`
+	Unknown     int64 `query:"count(*) filter (where am.advisory_type_id = 0)" gorm:"column:unknown"`
+	Unspecified int64 `query:"count(*) filter (where am.advisory_type_id = 1)" gorm:"column:unspecified"`
+	Other       int64 `query:"count(*) filter (where am.advisory_type_id = 2)" gorm:"column:other"`
+	Enhancement int64 `query:"count(*) filter (where am.advisory_type_id = 3)" gorm:"column:enhancement"`
+	Bugfix      int64 `query:"count(*) filter (where am.advisory_type_id = 4)" gorm:"column:bugfix"`
+	Security    int64 `query:"count(*) filter (where am.advisory_type_id = 5)" gorm:"column:security"`
```
MichaelMraka, line 52: Nice. This is correct except for Unknown, Unspecified and  Other :).
The thing is we care only about Security, Bugfix, Enhancement and anything else is counted under Other.
See e.g. metrics
https://github.com/RedHatInsights/patchman-engine/blob/master/vmaas_sync/metrics.go#L256

MichaelMraka (changes requested):

MichaelMraka:
Please remove the `fix(..)` from the first commit message we do not use this kind of annotation in patch.
Actually it will be better to squash the two commits together before merging.

MichaelMraka:
Please add some test for the sub totals.

michalslomczynski:
@MichaelMraka I added test to subtotals, what do you think?

manager/controllers/advisories_test.go
```diff
@@ -321,3 +321,25 @@ func TestAdvisoryTagsInMetadata(t *testing.T) {
...
+			bugfix++
+		case 3:
+			security++
+		default:
+			other++
+		}
+	}
+	st := output.Meta.SubTotals
+	assert.Equal(t, st["other"], other)
+	assert.Equal(t, st["enhancement"], enhancement)
```
MichaelMraka, line 345: Just a last nitpick: assert argument order is `expected`, `found` so in this case it should be
`assert.Equal(t, enhancement, st["enhancement"])`
and similarly for other asserts here.

MichaelMraka (changes requested):

manager/controllers/advisories_test.go
michalslomczynski: Very good catch, thanks!

PR #921, opened 2022-03-11
SPM-1404: Fix multiple SAP IDs

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


michalslomczynski:
After these changes, multiple SID ID do not overwrite itselves, but they stack up and are passed as an array to the database instead.

manager/controllers/systems_test.go
```diff
@@ -287,6 +287,16 @@ func TestSAPSystemMeta3(t *testing.T) {
 	assert.Equal(t, testMap, output.Meta.Filter)
 }
 
+func TestSAPSystemMeta4(t *testing.T) {
+	url := "/?filter[system_profile][sap_sids][in]=ABC&filter[system_profile][sap_sids][in]=GHI"
```
Josca, line 291: Do we really need to support this? Does it also work for another filters? Can not we we just use:
~~~
filter[system_profile][sap_sids][in]=ABC,GHI
~~~
?

Josca (changes requested):

manager/controllers/systems_test.go
michalslomczynski: I wanted to do this the way you just proposed, but unfortunately FE is already implemented this way `	https://console.stage.redhat.com/api/inventory/v1/hosts?staleness=fresh&staleness=stale_warning&=&registered_with=insights&filter[system_profile][mssql][version]=not_nil&filter[system_profile][sap_sids][]=EM1&filter[system_profile][sap_sids][]=EM2`

I would be more than happy to add support for this but this one gets the job done and we have to move on with other things AFAIK. What do you think about it Pepa?

Josca (approved):

PR #931, opened 2022-03-21
SPM-1326: Handle updateinfo in evaluator

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/evaluator/evaluate.go
+++ b/evaluator/evaluate.go
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	uploadTopic = "patchman.evaluator.upload"
-	recalcTopic = "patchman.evaluator.recalc"
+	uploadLabel = "upload"
+	recalcLabel = "recalc"
 )
 
 type SystemAdvisoryMap map[string]models.SystemAdvisories
@@ -100,7 +100,7 @@ func Evaluate(ctx context.Context, accountID int, inventoryID, accountName strin
 		return nil
 	}
 
-	system, vmaasData, err := evaluateInDatabase(ctx, accountID, inventoryID, accountName, requested)
+	system, vmaasData, err := evaluateInDatabase(ctx, accountID, inventoryID, accountName, requested, evaluationType)
 	if err != nil {
 		return errors.Wrap(err, "unable to evaluate in database")
 	}
@@ -117,7 +117,7 @@ func Evaluate(ctx context.Context, accountID int, inventoryID, accountName strin
 }
 
 func evaluateInDatabase(ctx context.Context, accountID int, inventoryID, accountName string,
-	requested *base.Rfc3339Timestamp) (*models.SystemPlatform, *vmaas.UpdatesV2Response, error) {
+	requested *base.Rfc3339Timestamp, evalType string) (*models.SystemPlatform, *vmaas.UpdatesV2Response, error) {
 	tx := database.Db.WithContext(base.Context).Begin()
 	// Don't allow requested TX to hang around locking the rows
 	defer tx.Rollback()
@@ -127,7 +127,7 @@ func evaluateInDatabase(ctx context.Context, accountID int, inventoryID, account
 		return nil, nil, nil
 	}
 
-	updatesData, err := getUpdatesData(ctx, tx, system)
+	updatesData, err := getUpdatesData(ctx, tx, system, evalType)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "unable to get updates data")
 	}
@@ -184,7 +184,8 @@ func evaluateWithVmaas(tx *gorm.DB, updatesData *vmaas.UpdatesV2Response,
 	return updatesData, nil
 }
 
-func getUpdatesData(ctx context.Context, tx *gorm.DB, system *models.SystemPlatform) (*vmaas.UpdatesV2Response, error) {
+func getUpdatesData(ctx context.Context, tx *gorm.DB, system *models.SystemPlatform, evalType string) (
+	*vmaas.UpdatesV2Response, error) {
 	var yumUpdates *vmaas.UpdatesV2Response
 	var yumErr error
 	if enableYumUpdatesEval {
@@ -195,7 +196,7 @@ func getUpdatesData(ctx context.Context, tx *gorm.DB, system *models.SystemPlatf
 		}
 
 		// in evaluator-upload return just return YumUpdates
-		if yumUpdates != nil && evalTopic == uploadTopic {
+		if yumUpdates != nil && evalType == uploadLabel {
 			return yumUpdates, nil
 		}
 	}
--- a/evaluator/evaluate_test.go
+++ b/evaluator/evaluate_test.go
@@ -102,7 +102,7 @@ func TestEvaluateYum(t *testing.T) {
 
 	mockWriter := mqueue.MockKafkaWriter{}
 	remediationsPublisher = &mockWriter
-	evalTopic = recalcTopic
+	evalLabel = recalcLabel
 
 	expectedAddedAdvisories := []string{"RH-1", "RH-2", "RHSA-2021:3801"}
 	expectedAdvisoryIDs := []int{1, 2, 14} // advisories expected to be paired to the system after evaluation


michalslomczynski:
I approached this by merging results of update lists from VMaaS and YumUpdates. Do you think it's the way to go without going into implementation details for now?

michalslomczynski (commented):
I will fix tests interference tomorrow.

michalslomczynski:
@MichaelMraka @Josca It should be good for first round of review. Sorry for pinging you too early guys :smile:


```
...
"bash-0:4.4.20-1.el8_4.x86_64": {
				"available_updates": [
					{
						"package": "bash-0:4.4.20-2.el8.x86_64",
						"repository": "rhel-8-for-x86_64-baseos-rpms",
						"basearch": "x86_64",
						"releasever": "8"
					}
				]
			}
...
+
...
"bash-0:4.4.21-1.el8_4.x86_64": {
				"available_updates": [
					{
						"package": "bash-0:4.4.21-2.el8.x86_64",
						"repository": "rhel-8-for-x86_64-baseos-rpms",
						"basearch": "x86_64",
						"releasever": "8"
					}
				]
			}
...
|
\/
"bash-0:4.4.20-1.el8_4.x86_64": {
				"available_updates": [
					{
						"package": "bash-0:4.4.20-2.el8.x86_64",
						"repository": "rhel-8-for-x86_64-baseos-rpms",
						"basearch": "x86_64",
						"releasever": "8"
					},
					{
						"package": "bash-0:4.4.21-2.el8.x86_64",
						"repository": "rhel-8-for-x86_64-baseos-rpms",
						"basearch": "x86_64",
						"releasever": "8"
					}
				]
			}
"bash-0:4.4.21-1.el8_4.x86_64": {
				"available_updates": [
					{
						"package": "bash-0:4.4.21-2.el8.x86_64",
						"repository": "rhel-8-for-x86_64-baseos-rpms",
						"basearch": "x86_64",
						"releasever": "8"
					}
				]
			}
```
@MichaelMraka We would have to parse NEVRAs to be able to handle this corner case, can we not handle that? Or do we want it and we accept a performance hit? It is O(n^2) algorithm in merge already so with regexes parsing it could get slooow.

base/utils/rpm.go
```diff
@@ -96,3 +97,32 @@ func (n Nevra) EVRAStringE(showEpoch bool) string {
 func (n Nevra) EVRAString() string {
 	return n.EVRAStringE(false)
 }
+
+func (n Nevra) IsLessVersion(other *Nevra) (bool, error) {
+	versionA := strings.Split(n.Version, ".")
```
MichaelMraka: Comparing NEVRAs is (way) more complex and it's already implemented in (system) rpmlib so we don't need to reinvent it.

I've pushed the fix into your branch.

conf/evaluator_recalc.env
```diff
@@ -2,3 +2,4 @@ EVAL_TOPIC=patchman.evaluator.recalc
 EVAL_LABEL=recalc
 PORT=8084
 CONSUMER_COUNT=8
+ENABLE_YUM_UPDATES_EVAL=true
```
MichaelMraka, line 5: It's true by default. We don't need to add it into test environment.

MichaelMraka (changes requested):

base/utils/rpm.go
michalslomczynski: Awesome! Much, much cleaner.

base/utils/vmaas.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, errors.New("release versions do not match")
+	}
+
+	if !RepoListsEqual(vmaasDataA.RepositoryList, vmaasDataB.RepositoryList) {
+		return nil, errors.New("repository lists differ")
+	}
+	if !ModuleListsEqual(vmaasDataA.ModulesList, vmaasDataB.ModulesList) {
+		return nil, errors.New("module lists differ")
+	}
+
```
MichaelMraka, line 26: I've discussed this (different repository or module list) with others.
First thing is the data come from the same upload so the should be consistent and even if not it would be better to use one of them as a primary source then abort whole evaluation and not return anything.

So I removed this extra check.

base/utils/vmaas.go
```diff
@@ -0,0 +1,158 @@
...
+			offset++
+			break
+		}
+		i++
+	}
+	if offset <= len(updatesB)-1 {
+		newUpdates = append(newUpdates, updatesB[offset:]...)
+	}
+	return &newUpdates, nil
+}
```
MichaelMraka, line 110: The merge loop works (according to the tests) but it's unnecessary hard to understand.
I've rewritten it into something cleaner (hopefully).

base/utils/vmaas.go
```diff
@@ -0,0 +1,158 @@
...
+		for _, moduleB := range *listB {
+			if moduleA.ModuleName == moduleB.ModuleName {
+				exists = true
+			}
+		}
+		if !exists {
+			return false
+		}
+	}
+
```
MichaelMraka, line 136: If the moduleB list is longer (a superset of moduleA) this function returns true. Which is not correct.

base/utils/vmaas.go
```diff
@@ -0,0 +1,158 @@
...
+		exists := false
+		for _, repoB := range *listB {
+			if repoB == repoA {
+				exists = true
+			}
+		}
+		if !exists {
+			return false
+		}
+	}
```
MichaelMraka, line 155: The same problem as in ModuleListEqual(). If listB is superset of listA, it returns true.

base/utils/vmaas_test.go
```diff
@@ -0,0 +1,245 @@
...
+	err := json.Unmarshal(updatesBash5, &listA)
+	assert.Nil(t, err)
+	err = json.Unmarshal(updatesBash6, &listB)
+	assert.Nil(t, err)
+
+	merged, err := mergeUpdates(listA, listB)
+	assert.Nil(t, err)
+
+	assert.Equal(t, 8, len(merged.GetAvailableUpdates()))
+	assert.Equal(t, true, updatesAscending(merged.GetAvailableUpdates()))
```
MichaelMraka, line 56: If you see a code repetition it's probably a good candidate to put it into function.

base/utils/vmaas_test.go
```diff
@@ -0,0 +1,245 @@
...
+	err := json.Unmarshal(kernel3101, &respA)
+	assert.Nil(t, err)
+	err = json.Unmarshal(kernel3102, &respB)
+	assert.Nil(t, err)
+
+	res, err := MergeVMaaSResponses(&respA, &respB)
+	assert.Nil(t, err)
+
+	updateList := (*res.UpdateList)[kernel310Pkg]
+	assert.Equal(t, 6, len(updateList.GetAvailableUpdates()))
```
MichaelMraka, line 192: There's a bug. Merging `kernel3101` and `kernel3102` should give you 7 different updates.

evaluator/evaluate.go
```diff
@@ -158,27 +174,86 @@ func evaluateWithVmaas(ctx context.Context, tx *gorm.DB, updatesReq *vmaas.Updat
...
+		return vmaasData, nil
 	}
+	if yumUpdates == nil && vmaasData == nil {
+		return nil, nil
+	}
+
+	// Return YumUpdates if it is uploader
+	if evalTopic == uploadTopic {
+		return yumUpdates, nil
+	}
```
MichaelMraka, line 205: This `if` should go before `getVmaasUpdates()` otherwise you unnecessarily call (and wait for and parse) vmaas and then just throw the response away.

evaluator/evaluate.go
```diff
@@ -158,27 +174,86 @@ func evaluateWithVmaas(ctx context.Context, tx *gorm.DB, updatesReq *vmaas.Updat
...
+	if vmaasErr != nil || vmaasData == nil {
+		return yumUpdates, nil
+	}
+	if yumErr != nil || yumUpdates == nil {
+		return vmaasData, nil
 	}
+	if yumUpdates == nil && vmaasData == nil {
+		return nil, nil
+	}
```
MichaelMraka, line 200: Having no yum updates is not an (hard) error. It may happen on RHEL6 which does not support it or in case there's and old insights client on the system.

base/utils/vmaas.go
MichaelMraka: Actually there's a bug in it, see my comment in the kernel test.

MichaelMraka:
I've pushed my fixes for mentioned issues.

base/utils/vmaas.go
michalslomczynski: Yes, that function was simple before speed optimization, definitely not after.

evaluator/evaluate.go
michalslomczynski: I think we only skip evaluation with VMaaS, no hard errors.

michalslomczynski: If both are nil

base/utils/vmaas_test.go
michalslomczynski: These versions are unique `10.1, 10.4, 10.7, 10.9, 11.0, 12.8`. I see 6.

base/utils/vmaas.go
michalslomczynski: Very nice, much cleaner!

base/utils/vmaas_test.go
michalslomczynski: @MichaelMraka do you confirm, or there indeed is something there?

michalslomczynski:
Great changes. Will you merge?

base/utils/vmaas_test.go
MichaelMraka: The  `10.9` version have different errata so it should be twice in the list. Updates are not completely equal.

michalslomczynski: I made up this data, do you think it will happen?

MichaelMraka: Yes, it actually can happen. E.g. java package is in more repos (one for base RHEL, other for layered products) and updates go to all of them but as a different errata.

base/utils/rpm.go
```diff
@@ -96,3 +98,11 @@ func (n Nevra) EVRAStringE(showEpoch bool) string {
 func (n Nevra) EVRAString() string {
 	return n.EVRAStringE(false)
 }
+
+func (n Nevra) Cmp(other *Nevra) int {
+	ret := strings.Compare(n.Name, other.Name)
+	if ret != 0 {
+		return ret
+	}
```
psegedy, line 106: I'm just curious, what is the purpose of comparing rpm names?

psegedy, line 106: I'm just curious, what is the purpose of comparing rpm names?

MichaelMraka: The aim is to sort NEVRAs, rpmlib can only compare EVRA, so the first thing you sort/compare is name, then EVRA.

evaluator/evaluate.go
```diff
@@ -165,27 +181,84 @@ func evaluateWithVmaas(ctx context.Context, tx *gorm.DB, updatesReq *vmaas.Updat
...
+		yumUpdates, yumErr = tryGetYumUpdates(system)
+		if yumErr != nil {
+			// ignore broken yum updates
+			utils.Log("Can't parse yum_updates").Warn(yumErr.Error())
+		}
+
+		// in evaluator-upload return just return YumUpdates
+		if yumUpdates != nil && evalTopic == uploadTopic {
+			return yumUpdates, nil
+		}
```
psegedy, line 200: Just for clarification.
- evaluator-upload should evaluate systems only by yum_updates and it should not call vmaas.
- evaluator-recalc should use yum_updates and vmaas response and merge updates
is that correct?

MichaelMraka: Exactly.

evaluator/evaluate.go
```diff
@@ -21,6 +21,11 @@ import (
 	"gorm.io/gorm/clause"
 )
 
+const (
+	uploadTopic = "patchman.evaluator.upload"
+	recalcTopic = "patchman.evaluator.recalc"
+)
+
```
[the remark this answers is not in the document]
MichaelMraka: It's tricky, we don't know what the name of topic will be... but we already have EVAL_LABEL to distinguish upload and recalc so I switch to checking label.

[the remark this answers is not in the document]
psegedy: thanks

PR #957, opened 2022-05-11
SPM-1484: Add instant advisory notifications

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/notification/notification.go
+++ b/base/notification/notification.go
@@ -1,0 +1,80 @@
+package notification
+
+import (
+	"strconv"
+	"time"
+)
+
+const (
+	Version     = "v1.1.0"
+	Bundle      = "rhel"
+	Application = "patch"
+)
+
+type Context struct {
+	InventoryID string `json:"inventory_id"`
+}
+
+type Event struct {
+	// Future-proof. Needs to be there and empty for now.
+	Metadata interface{} `json:"metadata,omitempty"`
+	// Your application payload.
+	// All the data required by the app to compose the various messages (Email, webhook…​) after transformation.
+	Payload interface{} `json:"payload,omitempty"`
+}
+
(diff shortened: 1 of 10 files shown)


michalslomczynski:
Maybe `last_updated` column can be dropped. Will be sure after aggregation in notifications-backend.

base/models/models.go
```diff
@@ -139,6 +139,14 @@ type AdvisoryType struct {
 	Preference int
 }
 
+var AdvisoryTypes = map[int]string{
+	0: "unknown",
+	1: "enhancement",
+	2: "bugfix",
+	3: "security",
+	4: "unspecified",
+}
```
MichaelMraka, line 148: You should read this enumeration from database not hardcode it.

MichaelMraka (changes requested):

base/models/models.go
michalslomczynski: I wanted to discuss this one with you. I've seen we read those values for each request on some endpoint. These values should not change during runtime so wouldn't it be better if we used singleton to read from instead? In similar way as config vars. What do you think @MichaelMraka ?

michalslomczynski: And to be clear reading these from DB for each notification push is not a problem at all, so we can discuss about it later as well if you prefer :wink:

michalslomczynski: Already changed for querying DB for now and lets talk about it later in improvement scope.

MichaelMraka (approved):

PR #960, opened 2022-05-13
SPM-1482: org_id migration

psegedy:
- add org_id column
- create new accounts using both account_number and org_id
- assign org_id to existing accounts
- handle case if account_number is deleted from model
- make sure we won't overwrite org_id if we have more records with null account_number and receive identity header with empty account_number
- create accounts using account_number from kafka messages because they are not sending org_id in the msg yet

~~todo:~~
~~- build AccountIDCache correctly if we receive only account_number (from kafka msg)~~
~~- make sure we won't leak data to wrong user. It can happen that we will have same ID in cache when some account_number of one user is the same as org_id of other user~~ should not happen, we are not using findAccount in listener

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


manager/middlewares/authentication.go
```diff
@@ -20,38 +19,69 @@ var AccountIDCache = struct {
...
 	}
 	if rhAccount.ID != 0 {
 		return rhAccount.ID, nil
 	}
-	err = database.OnConflictUpdate(database.Db, "name", "name").Create(&rhAccount).Error
+
+	// OrgID not found, try to find account by account_number and update org_id or create new record
+	if rhAccount.OrgID == "" {
+		// kafka msg without OrgID, create account by AccountNumber
+		err = database.OnConflictUpdate(database.Db, "name", "name").Select("name").Create(&rhAccount).Error
```
MichaelMraka, line 49: Isn't `.Select("name")` redundant here?

manager/middlewares/authentication.go
```diff
@@ -20,38 +19,69 @@ var AccountIDCache = struct {
...
+		if err != nil {
+			utils.Log("err", err, "name", rhAccount.Name).Warn("Error creating account")
+		}
+		return rhAccount.ID, err
+	}
+	// create new rhAccount with OrgID
+	if rhAccount.Name == "" {
+		// avoid updating other account with name=null if AccountNumber=""
+		err = database.OnConflictUpdate(database.Db, "org_id", "org_id").Select("org_id").Create(&rhAccount).Error
```
MichaelMraka, line 58: And similarly `.Select("org_id")` redundant here?

MichaelMraka:
Please rebase it on top of current master.

manager/middlewares/authentication.go
psegedy: likewise for Select(org_id)

manager/middlewares/authentication.go
MichaelMraka: Ok, I understand.

MichaelMraka (changes requested):

PR #982, opened 2022-06-20
SPM-1484: Use RH account name instead of ID

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/notification/notification.go
+++ b/base/notification/notification.go
@@ -1,7 +1,6 @@
 package notification
 
 import (
-	"strconv"
 	"time"
 )
 
@@ -57,6 +56,8 @@ type Notification struct {
 	// Recipients settings - Applications can add extra email recipients by adding entries to this array.
 	// This setting extends whatever the Administrators configured in their Notifications settings (since v1.1.0).
 	Recipients []Recipient `json:"recipients,omitempty"`
+	// org_id as future replacement of the account_id.
+	OrgID string `json:"org_id,omitempty"`
 }
 
 type Advisory struct {
@@ -65,7 +66,7 @@ type Advisory struct {
 	Synopsis     string `json:"synopsis"`
 }
 
-func MakeNotification(accountID int, inventoryID string, eventType string, events []Event) *Notification {
+func MakeNotification(rhAccountName string, inventoryID string, eventType string, events []Event) *Notification {
 	return &Notification{
 		Version:     Version,
 		Bundle:      Bundle,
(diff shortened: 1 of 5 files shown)


base/notification/notification.go
```diff
@@ -57,6 +56,8 @@ type Notification struct {
 	// Recipients settings - Applications can add extra email recipients by adding entries to this array.
 	// This setting extends whatever the Administrators configured in their Notifications settings (since v1.1.0).
 	Recipients []Recipient `json:"recipients,omitempty"`
+	// org_id as future replacement of the account_id.
+	OrgID string `json:"org_id,omitempty"`
```
psegedy, line 60: why can't we use `OrgID`? If notifications support it, you can use it right away, we should be using OrgID everywhere where it is possible.

evaluator/notifications.go
```diff
@@ -35,21 +36,30 @@ func getNotificationAdvisories(newAdvisories SystemAdvisoryMap) []ntf.Advisory {
...
 
-	advisories := getNotificationAdvisories(newAdvisories)
+	var rhAccountName string
+	err := tx.Table("rh_account").
+		Select("name").
+		Where("id = ?", accountID).
+		Take(&rhAccountName).Error
+	if err != nil {
+		return errors.Wrap(err, "querying database for RH account name failed")
+	}
```
psegedy, line 52: you can also get Account Number or OrgID from PlatformEvent kafka message. Just pass it through `Evaluate` in `evaluate.go` (and other functions along the way). Ofc, if that makes sense to you.

base/notification/notification.go
michalslomczynski: Notifications REST API accepted integer only and for string raised `OrgID must be greater than 0` error, while API spec clearly said it wants string. Today REST API returns 302 moved permanently, so they are doing something with it. I am fine with passing it but my priority is to test notifications now and it can cause some delay if we would need to revert it.

michalslomczynski: I would pass it after testing to be safe :slightly_smiling_face:

evaluator/notifications.go
michalslomczynski: I did not know it's in PE, thanks.

base/notification/notification.go
psegedy: sure, thank you for explanation

psegedy (approved):
LGTM, let's wait for tests in PR check, last time there was a 500 from Inventory api

PR #986, opened 2022-06-22
SPM-1447: run evaluation in goroutines

psegedy:
It looks like it's helping, but just a bit. Recalc time went down from ~1h to ~40min (once I got 11 min, so I'm not sure...) when I used 50 goroutines in ephemeral env. IMO we need to see evals/sec metrics in some other environments.

The PR runs `Evaluate` function in Goroutines limited by MAX_EVAL_GOROUTINES env var.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


michalslomczynski:
I am curious results of benchmark in <CPU_CORES, 1000> range of goroutines.

psegedy:
well, I tried that only for recalc and it's not really easy to benchmark. You need to generate a lot of data, you need data synced in vmaas, then you need to trigger recalc which sometimes runs 11min and other times a lot longer... In the end its still on generated data and it does not really be close to reality. Maybe we could benchmark it with uploads, but again we would need thousands of systems with different packages, repos, and uploaded to different accounts to really hit the eval in batches (1 kafka message can contain multiple systems).

After all, we can't run very high number of goroutines, we are limited by memory and vmaas cannot send too many requests to vmaas.

The main issue these goroutines should solve is when you get a kafka message with a lot of systems (default max is 4000). It looks like we are seeing this scenario in prod with recalc messages. So, instead of processing these messages from 1 kafka message sequentially, we can try to evaluate them concurrently (but still we cannot halt vmaas) which should help.

MichaelMraka (approved):

psegedy:
🙏

PR #1018, opened 2022-07-11
use initContainer for db migration

psegedy:
Initially I wanted to use k8s job which runs migrations and add initContainer to all deployements to wait till migration is ready. However, using plain job suffers from clowder usage - it doesn't have the correct config and can't pull the image and ClowdJob triggered by ClowdJobInvoction can't be run if the app is not running so it can't be used for migrations.

The implementation is basically the same as in Vulnerability
1. manager pod has initContainer which runs migrations
2. to avoid concurrency we are using pg_advisory_lock
3. if the migration is not successful, init container panics which _should_ result in failure and new manager won't be deployed
4. other pods (evaluators, listeners,...) will start immediately and they may be in CrashLoop state until migrations are not successful

It is not the ideal solution, any recommendations?

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/database_admin/entrypoint.sh
+++ b/database_admin/entrypoint.sh
@@ -6,5 +6,3 @@ MIGRATION_FILES=file://./database_admin/migrations
 
 echo "Running in $(pwd) as $(id)"
 ${GORUN:+go run} ./main${GORUN:+.go} migrate $MIGRATION_FILES
-
-exec sleep infinity
--- a/database_admin/migrate.go
+++ b/database_admin/migrate.go
@@ -68,8 +68,8 @@ func MigrateUp(conn database.Driver, sourceURL string) {
 	}
 
 	if err != nil {
-		// Don't panic on error, Log and keep the container running so we can diagnose it
 		fmt.Fprintf(os.Stderr, "Error upgrading the database: %v", err.Error())
+		panic(err)
 	}
 }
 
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -16,6 +16,11 @@ RUN dnf module -y enable postgresql:12 && \
     dnf install -y go-toolset postgresql git-core diffutils rpm-devel && \
     ln -s /usr/libexec/platform-python /usr/bin/python3
 
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz \
+    -o oc.tar.gz && \
+    tar -C /usr/bin/ -xvzf oc.tar.gz oc && \
+    rm -f oc.tar.gz
+
```
MichaelMraka, line 23: On RHEL it's better to install `oc` from rpm:
`dnf install -y openshift-clients --enablerepo=rhocp-4.6-for-rhel-8-x86_64-rpms`

psegedy: I'm planning to try that as well but I wasn't sure if we have correct subscriptions on ci.ext jenkins nodes, so for now I used the tarball

MichaelMraka (approved):

PR #1022, opened 2022-07-12
SPM-1559: Refactor manager/controllers tests

Mischulee:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


manager/controllers/common_test.go
```diff
@@ -0,0 +1,72 @@
...
+	return httptest.NewRecorder(), req
+}
+
+// Check if additional header
+func CheckHeader(req *http.Request, contentType *string) {
+	if contentType != nil {
+		contentTypeVal := *contentType
+		req.Header.Add("Accept", contentTypeVal)
+	}
+}
```
psegedy, line 30: I have a few suggestions. 

- it doesn't need to be exported (if I didn't miss that), so it can have `lowerCase` name
- `CheckHeader` is a little misleading name since it does not actually checks anything. It adds `Accept` header. Something like `acceptContentType` or `addAcceptHeader` would be more clear in my opinion
- `contentType` can be `string` it is not neccessary to use `*string`, you will just check for `""` instead of `nil`. It will simplify some calls to `CreateRequest` where you would not need to define a variable holding content type value like this:
```go
// i've seen something like this in some test above
ct := "text/html"
CreateRequest(arg1, arg2, etc, &ct)
// you could call it as if you used pointer
CreateRequest(arg1, arg2, etc, "text/html")
```

```suggestion
func (req *http.Request) acceptContentType(ct string) {
	if ct != "" {
		req.Header.Add("Accept", ct)
	}
}
```

manager/controllers/common_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+)
+
+// Content type variables
+var contentTypeCSV = "text/csv"
+var contentTypeJSON = "application/json"
+
+// Init request
+func PrepareRequest(method string, url string, body io.Reader) (w *httptest.ResponseRecorder, req *http.Request) {
```
psegedy, line 19: doesn't have to be exported 
```suggestion
func prepareRequest(method, url string, body io.Reader) (w *httptest.ResponseRecorder, req *http.Request) {
```

manager/controllers/common_test.go
```diff
@@ -0,0 +1,72 @@
...
+func CreateRequestRouterWithAccount(method string, url string, body io.Reader, contentType *string,
+	handler gin.HandlerFunc, routerPath string, routerAccount int) (w *httptest.ResponseRecorder) {
+	w, req := PrepareRequest(method, url, body)
+	CheckHeader(req, contentType)
+	core.InitRouterWithAccount(handler, routerPath, routerAccount).ServeHTTP(w, req)
+	return w
+}
+
+// Check status and parse response body
+func ParseResponse(t *testing.T, w *httptest.ResponseRecorder, expectedStatus int, output interface{}) {
```
psegedy, line 69: I'd change the func name to something more clear, what do you think about `CheckResponse`?

manager/controllers/common_test.go
```diff
@@ -0,0 +1,72 @@
...
+		contentTypeVal := *contentType
+		req.Header.Add("Accept", contentTypeVal)
+	}
+}
+
+// Create simple request
+func CreateRequest(method string, url string, body io.Reader, contentType *string, handler gin.HandlerFunc) (
+	w *httptest.ResponseRecorder) {
+	w, req := PrepareRequest(method, url, body)
+	CheckHeader(req, contentType)
```
psegedy, line 36: what do you think about using `CheckHeader` inside the `PrepareRequest` function?

psegedy (changes requested):
It looks great, thank you for making tests more readable and maintainable. I just left some comments to make it a bit clearer.
Sorry it took so long to review it

psegedy:
and rebase it, please. There has been a change in `baseline_create_test` in the meantime

manager/controllers/common_test.go
Mischulee: You are right, this looks way better and cleaner. I will change it.

manager/controllers/common_test.go
Mischulee: I agree.

manager/controllers/common_test.go
Mischulee: ```CheckResponse``` seems like a more suitable name.

manager/controllers/common_test.go
Mischulee: Good idea, I will add that.

PR #1039, opened 2022-07-28
Minor logic improvements

michalslomczynski:
- First one is always true
- Second one is never nil

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/database/database.go
+++ b/base/database/database.go
@@ -36,7 +36,7 @@ func OnConflictDoUpdateExpr(db *gorm.DB, keys []string, updateExprs ...UpExpr) *
 	for _, key := range keys {
 		confilctColumns = append(confilctColumns, clause.Column{Name: key})
 	}
-	if updateColsValues != nil {
+	if len(updateColsValues) > 0 {
 		return db.Clauses(clause.OnConflict{
 			Columns:   confilctColumns,
 			DoUpdates: clause.Assignments(updateColsValues),
--- a/tasks/vmaas_sync/advisory_sync.go
+++ b/tasks/vmaas_sync/advisory_sync.go
@@ -224,7 +224,7 @@ func storeAdvisories(data map[string]vmaas.ErrataResponseErrataList) error {
 		return errors.WithMessage(err, "Parsing advisories")
 	}
 
-	if advisories == nil || len(advisories) == 0 {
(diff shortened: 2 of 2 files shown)


vmaas_sync/advisory_sync.go
```diff
@@ -223,7 +223,7 @@ func storeAdvisories(data map[string]vmaas.ErrataResponseErrataList) error {
 		return errors.WithMessage(err, "Parsing advisories")
 	}
 
-	if advisories == nil || len(advisories) == 0 {
```
psegedy, line 226: The change looks good, based on return type, the parseAdvisories should never return nil.

However, this whole file does not look good, `parseAdvisories` returns slice and error but when there is an error in `parseAdvisories` it returns `nil, err` which does not match the type. First of all, we need to turn on linter for this file and then we can make such changes

MichaelMraka: As for the change - both nil and empty slice have` len() == 0` that's not a problem.

As for returning `nil` in `parseAdvisories()` - it's also ok, slice can have nil value (uninitialized slice) and in 99% cases you don't have to take special care about it (you can append to and iterate over it just fine).

base/database/database.go
```diff
@@ -36,11 +36,8 @@ func OnConflictDoUpdateExpr(db *gorm.DB, keys []string, updateExprs ...UpExpr) *
...
-		return db.Clauses(clause.OnConflict{
-			Columns:   confilctColumns,
-			DoUpdates: clause.Assignments(updateColsValues),
-		})
-	}
-	return db
+	return db.Clauses(clause.OnConflict{
+		Columns:   confilctColumns,
+		DoUpdates: clause.Assignments(updateColsValues),
+	})
```
MichaelMraka, line 42: I understand `updateColsValues` is never `nil` beacuse it's created as an empty map.
But should not the condition be changed to `len(updateColsValues) > 0` to keep the original logic/intent?

MichaelMraka (approved):

PR #1061, opened 2022-08-17
SPM-1705: Refactor vmaas-sync recalc event

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/database/testing.go
+++ b/base/database/testing.go
@@ -446,3 +446,8 @@ func CheckBaselineDeleted(t *testing.T, baselineID int) {
 	assert.Nil(t, Db.Model(&models.SystemPlatform{}).Where("baseline_id = ?", baselineID).Count(&cntSystems).Error)
 	assert.Equal(t, 0, int(cntSystems))
 }
+
+func GetAllSystems(t *testing.T) (systems []*models.SystemPlatform) {
+	assert.Nil(t, Db.Model(&models.SystemPlatform{}).Order("rh_account_id").Scan(&systems).Error)
+	return systems
+}
--- a/base/mqueue/platform_event.go
+++ b/base/mqueue/platform_event.go
@@ -21,25 +21,19 @@ type PlatformEvent struct {
 	RequestIDs  []string                `json:"request_ids,omitempty"`
 }
 
-type InventoryAID struct {
-	InventoryID string
-	RhAccountID int
-	OrgID       *string
-}
-
 type EvalData struct {
-	RhAccountID int
 	InventoryID string
+	RhAccountID int
 	RequestID   string
 	OrgID       *string
 }
 
 type PlatformEvents []PlatformEvent
-type InventoryAIDs []InventoryAID
 type EvalDataSlice []EvalData
 
-type groupedData map[int][]string
-type orgIDMap map[int]*string
+type accountInventories map[int][]string
+type accountRequests map[int][]string
+type orgIDs map[int]*string
 
 func (event *PlatformEvent) createKafkaMessage() (KafkaMessage, error) {
 	data, err := json.Marshal(event) //nolint:gosec
@@ -75,7 +69,7 @@ func writePlatformEvents(ctx context.Context, w Writer, events ...PlatformEvent)
 	return w.WriteMessages(ctx, msgs...)
 }
 
-func batchSize(grouped groupedData) int {
+func batchSize(grouped map[int][]string) int {
 	// compute how many batches we will create
 	var batches = 0
 	for _, ev := range grouped {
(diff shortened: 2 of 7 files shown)


michalslomczynski:
Regarding refactor:
- Not a strong preference against previous variable names
- Very strong preference to get rid of this `groupedData` struct and `data` variable names.

psegedy:
I haven't had time to look at it properly. Wouldn't it be possible to completely remove everything related to `accountInventories` and use only `EvalDataSlice` struct?

base/mqueue/platform_event.go
```diff
@@ -83,46 +85,51 @@ func batchSize(grouped groupedData) int {
 	return batches
 }
 
-func (data *InventoryAIDs) groupData() (int, *groupedData) {
+func (accs *InventoryAccounts) getAccountInventories() (int, *accountInventories, *orgIDs) {
```
psegedy, line 88: there was an intention of having the same API/function name for multiple receiver types, no strong preference tho

base/mqueue/platform_event.go
```diff
@@ -131,21 +138,14 @@ func (data *InventoryAIDs) WriteEvents(ctx context.Context, w Writer) error {
...
-	groupedSysVal := groupedData{}
-	groupedReqIDVal := groupedData{}
-	if groupedSys != nil {
-		groupedSysVal = *groupedSys
-	}
-	if groupedReqID != nil {
-		groupedReqIDVal = *groupedReqID
-	}
+func (evals *EvalDataSlice) WriteEvents(ctx context.Context, w Writer) error {
+	// invs, reqs, orgIDs never nil
```
psegedy, line 142: what if someone will return nil in `getAccountEvalData` in the future? It's still a valid return value.
Can't we make `reqs=nil` and use EvalDataSlice instead of the InventoryAccounts? `reqs`'s the only difference in those 2 `WriteEvents` functions. Just a thought

tasks/vmaas_sync/repo_based.go
```diff
@@ -38,18 +38,19 @@ func getCurrentRepoBasedInventoryIDs() ([]mqueue.InventoryAID, error) {
...
 	}
 
 	err := database.Db.Table("system_repo sr").
 		Joins("JOIN repo ON repo.id = sr.repo_id").
 		Joins("JOIN system_platform sp ON  sp.rh_account_id = sr.rh_account_id AND sp.id = sr.system_id").
+		Joins("JOIN rh_account ra ON ra.id = sp.rh_account_id").
 		Where("repo.name IN (?)", repos).
 		Order("sp.rh_account_id").
-		Select("distinct sp.inventory_id, sp.rh_account_id").
+		Select("distinct sp.inventory_id, sp.rh_account_id, ra.org_id").
```
psegedy, line 53: I see it was there before, I just wonder why the `distinct` is needed

base/mqueue/platform_event.go
michalslomczynski: I refactored this because it was mixed approach of assuming dereferencing is one time safe and other is not, while they all are the same, never nil pointers, so lack of uniformity.

Now it's better because if you handle nil pointer inside getter function, you don't repeat itself every time you use it and do not obfuscate other functions. It's very popular to handle nil pointers in getters and return default value if it's nil - empty struct in this case.

If one changes getter, he ought to check it's receivers.

Mayve EvalDataSlice can be reused indeed. I'll take a look.

base/mqueue/platform_event.go
michalslomczynski: I see the reason for `WriteEvents` same function name, it's part of `MessageData` interface, but I have no clue for these getters. What is the reason?

It's better to use names that suggest what variable holds. These new names may be improved, but `InventoryAccounts` is still lesser set than `groupedData`, that can refer to almost anything :smile:

tasks/vmaas_sync/repo_based.go
michalslomczynski: Let's ping pepa :smile:

michalslomczynski: Mim authored this one, he will know. Looks like safety check or corner case handle. Do you think it can be a problem?

base/mqueue/platform_event.go
michalslomczynski: Slice in Go is just
```
type SliceHeader struct {
	Data [uintptr](https://pkg.go.dev/builtin#uintptr)
	Len  [int](https://pkg.go.dev/builtin#int)
	Cap  [int](https://pkg.go.dev/builtin#int)
}
```
https://pkg.go.dev/reflect#SliceHeader
So all the heavy stuff is pointer anyway here, so probably it will be faster to return from these getters as value, and we will get rid of nil pointer issue :slightly_smiling_face:

michalslomczynski: It's the same performance
```
By value
    1672            646303 ns/op
By reference
    1602            658101 ns/op
```
Let's just drop pointers from getters.

psegedy: I agree with passing by values, that's a good point. Please try to look if it is possible to use only `EvalDataSlice`

tasks/vmaas_sync/repo_based.go
psegedy: not a problem, just curious. Distinct might be slowing this down a bit

michalslomczynski:
@psegedy I replaced it in last commit, but removed `EvalData`.

I personally prefer `InventoryAccount` over `EvalData` since you do not know what `EvalData` is without checking it's implementation. It is not improved readability if you have to do this IMO. `InventoryAccount` name is more specific. This struct just holds the account of the inventory, and the name suggest exactly that, simple.

tasks/vmaas_sync/repo_based.go
michalslomczynski: Better call @MichaelMraka.

base/database/testing.go
```diff
@@ -442,3 +442,8 @@ func CheckBaselineDeleted(t *testing.T, baselineID int) {
 	assert.Nil(t, Db.Model(&models.SystemPlatform{}).Where("baseline_id = ?", baselineID).Count(&cntSystems).Error)
 	assert.Equal(t, 0, int(cntSystems))
 }
+
+func GetAllSystems(t *testing.T) (systems []*models.SystemPlatform) {
+	assert.Nil(t, Db.Model(&models.SystemPlatform{}).Select("*").Order("rh_account_id").Scan(&systems).Error)
+	return systems
```
MichaelMraka, line 448: Select() is redundant, `Db.Model(&models.SystemPlatform{}).Order("rh_account_id").Scan(&systems).Error` is enough.

MichaelMraka:
@michalslomczynski, I have hard times to understand this PR. Can you please split it into
- necessary orgId changes,
- independent refactoring PR?

michalslomczynski:
Necessary OrgID changes:
[SPM-1705: Add OrgID field to eval recalc messages](https://github.com/RedHatInsights/patchman-engine/pull/1061/commits/7e3101f70d7dd71b82b3ed349f01a640da4c9d78)

michalslomczynski:
To sum up refactor in case it's hard to track what was changed:
- Removed improper use of Go types
- Removed improper use of pointers
~~- Renamed inventoryAID/inventoryAIDs -> InventoryAccount/InventoryAccounts~~ Reverted this in favor of `EvalDataSlice` Patrick suggested
- Removed one of `MessageData` interface implementation because it was redundant.

tasks/vmaas_sync/repo_based.go
michalslomczynski: @psegedy This should be fine.

michalslomczynski:
@psegedy I am fine with EvalDataSlice and EvalData names, but we have this weird old variables all over the place, eg. `inventoryAIDs = EvalDataSlice{...}`.

psegedy:
@michalslomczynski not a big deal IMO and it would require renaming also some functions. Maybe something like `inventoryAccounts` would be more appropriate now. It used to contain only inventoryID and accountID, but now it also has orgID. `inventoryAIDs` might not be very descriptive, I guess it should stand for `inventory account ids` which is technically still true with orgIDs, I'd let it be as it is.

MichaelMraka (approved):

psegedy (approved):

PR #1169, opened 2023-02-08
make map allocation consistent in the app

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/evaluator/remediations.go
+++ b/evaluator/remediations.go
@@ -42,8 +42,9 @@ func createRemediationsStateMsg(id string, response *vmaas.UpdatesV2Response) *R
 }
 
 func getReportedAdvisories(vmaasData *vmaas.UpdatesV2Response) map[string]bool {
-	advisories := map[string]bool{}
-	for _, updates := range vmaasData.GetUpdateList() {
+	updateList := vmaasData.GetUpdateList()
+	advisories := make(map[string]bool, len(updateList))
+	for _, updates := range updateList {
 		for _, u := range updates.GetAvailableUpdates() {
 			advisories[u.GetErratum()] = true
 		}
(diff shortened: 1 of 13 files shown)


evaluator/remediations.go
```diff
@@ -42,8 +42,9 @@ func createRemediationsStateMsg(id string, response *vmaas.UpdatesV2Response) *R
 }
 
 func getReportedAdvisories(vmaasData *vmaas.UpdatesV2Response) map[string]bool {
-	advisories := map[string]bool{}
-	for _, updates := range vmaasData.GetUpdateList() {
+	updateList := vmaasData.GetUpdateList()
+	advisories := make(map[string]bool, len(updateList))
+	for _, updates := range updateList {
```
psegedy, line 47: not sure if this is actually better or worse because we are allocating len(updateList) but there is updateList*updates.GetAvailableUpdates items in the slice. I think golang allocates 2*current_size when it needs to extend a map. So in this case it might actually allocate more space compared to `make(map[string]bool, 0)`

psegedy (approved):
I'm not sure about 1 allocation I commented, otherwise LGTM

evaluator/remediations.go
MichaelMraka: Ah, I see. On the other hand let's assume typical `updates.GetAvailableUpdates()` return 2-3 items so in the end after first extension of map it will be big enough to hold all keys (in most cases :).

PR #1171, opened 2023-02-14
SPM-1880: add v3 to /system/advisories

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


manager/controllers/advisories.go
```diff
@@ -32,38 +33,70 @@ type AdvisoryMetaTotalHelper struct {
...
+	AdvisoriesDBLookupCommon
+	AdvisoryItemAttributesV3
 }
 
 // nolint: lll
-type AdvisoryItemAttributes struct {
+type AdvisoryItemAttributesV2 struct {
 	SystemAdvisoryItemAttributes
-	ApplicableSystems int `json:"applicable_systems" query:"COALESCE(aad.systems_installable, 0)" csv:"applicable_systems" gorm:"column:applicable_systems"`
+	InstallableSystems int `json:"installable_systems" query:"COALESCE(aad.systems_installable, 0)" csv:"installable_systems" gorm:"column:installable_systems"`
```
psegedy, line 55: I think we should keep `applicable_systems` name for `v2` api, otherwise it can break the current UI (and 1 smoke test is already failing because of that)

IMO something like the following should work, use `aad.systems_installable` column but expose it as `applicable_systems` in api response
```go
type AdvisoryItemAttributesV2 struct {
	SystemAdvisoryItemAttributes
	ApplicableSystems int `json:"applicable_systems" query:"COALESCE(aad.systems_installable, 0)" csv:"applicable_systems" gorm:"column:applicable_systems"`
}

type AdvisoryItemAttributesV3 struct {
	SystemAdvisoryItemAttributes
	InstallableSystems int `json:"installable_systems" query:"COALESCE(aad.systems_installable, 0)" csv:"installable_systems" gorm:"column:installable_systems"`
	ApplicableSystems int `json:"applicable_systems" query:"COALESCE(aad.systems_applicable, 0)" csv:"applicable_systems" gorm:"column:applicable_systems"`
}
```

psegedy: hmm or maybe you would need also something like `AdvisoryItemAttributesAll` which would consist of common attributes, V2Only attributes, and V3Only attributes. Because with my proposed solution above,  `advisoryItemV3toV2` won't work :(

psegedy (changes requested):

manager/controllers/advisories.go
MichaelMraka: That's right, good catch.

PR #1183, opened 2023-03-02
fix: ensure openssl is installed in ubi8-micro

SteveHNH:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,10 @@ RUN dnf module -y enable postgresql:12 && \
     dnf install -y go-toolset postgresql git-core diffutils rpm-devel && \
     ln -s /usr/libexec/platform-python /usr/bin/python3
 
+RUN mkdir -p /mnt/rootfs && \
+    dnf install --installroot /mnt/rootfs --releasever 8 --setopt install_weak_deps=false --nodocs -y openssl && \
+    rm -rf /mnt/rootfs/var/cache/*
+
 ENV GOPATH=/go \
     GO111MODULE=on \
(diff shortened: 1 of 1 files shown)


SteveHNH:
Attempted to just copy the openssl binary and libraries but was still unable to overcome the error. Went back to the original idea of doing a dnf install against a temp rootfs and then copying that into the buildimg. Picked up that method from this dockerfile: https://github.com/fatherlinux/ubi-micro/blob/master/ubi8-micro-openssl

It gets the pod to run properly, but if someone wants to try using the ldd method, they're welcome to it. I think I've taken it as far as I can on my own with limited knowledge of the app itself.

MichaelMraka:
Hello @SteveHNH , I think it's likely some configuration in /etc/ what's missing.
I can try it, is there a way how to check the image works in fedramp?

SteveHNH:
@MichaelMraka sure! If you want to take a crack at it and ping me here or on slack, I can run the image in FR and see if it works there. As long as the pr builds an image for quay, we can use it.

syncrou:
@MichaelMraka due to the tight deadline on this. Do you see a real issue with getting this PR merged now since @SteveHNH know's it works in our environment?

It would be great for us to be able to check another thing off our list so we can keep moving forward.

PR #1277, opened 2023-08-01
RHINENG-1197: fix meta links 

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/manager/controllers/utils.go
+++ b/manager/controllers/utils.go
@@ -96,15 +96,15 @@ type NestedFilterMap map[string]string
 var nestedFilters = NestedFilterMap{
 	"group_name":                                  "group_name",
 	"group_name][in":                              "group_name", // obsoleted, backward compatible
-	"system_profile][sap_system":                  "sap_system",
-	"system_profile][sap_sids":                    "sap_sids",
-	"system_profile][sap_sids][":                  "sap_sids", // obsoleted, backward compatible
-	"system_profile][sap_sids][in":                "sap_sids", // obsoleted, backward compatible
-	"system_profile][sap_sids][in][":              "sap_sids", // obsoleted, backward compatible
-	"system_profile][ansible":                     "ansible",
-	"system_profile][ansible][controller_version": "ansible->controller_version",
-	"system_profile][mssql":                       "mssql",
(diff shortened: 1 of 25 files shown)


manager/controllers/utils.go
```diff
@@ -29,15 +29,6 @@ var tagRegex = regexp.MustCompile(`([^/=]+)/([^/=]+)(=([^/=]+))?`)
...
 
-var validSystemProfileFilters = map[string]bool{
-	"sap_sids":                    true,
-	"sap_system":                  true,
-	"mssql":                       true,
-	"mssql->version":              true,
-	"ansible":                     true,
-	"ansible->controller_version": true,
-}
-
```
psegedy, line 40: don't we need this map to allow only certain keywords for filter? We should be able to return 400 if user tries something like `filter[system_profile][asdf]=asdf` but I don't know if we can do it with the old code or not

psegedy (commented):
we should make sure we don't break any checks for invalid filters if we have them now, I'd rely on unit tests and QE to provide an answer whether the functionality is correct or not

if it doesn't break any current checks for invalid filters, then LGTM

PR checks are broken until Brandon and app-sre make necessary changes

manager/controllers/utils.go
MichaelMraka: It's checked against` nestedFilters` map
https://github.com/RedHatInsights/patchman-engine/pull/1277/files/fc656574644903bfb6a2aaba18ee18b3e12f91c2..47c4a56c8ea79e937dab87416df48749c0ba456d#diff-dca21f2371a54b52831672e4a621006fc040a2986ce17b3bfd408b8b3cc2452bL105-L117

MichaelMraka:
There's couple of invalid filter tests, e.g.
TestSystemsFilterNotExisting
TestSystemsIDsFilterNotExisting

psegedy (approved):

PR #1361, opened 2024-01-05
RHINENG-6806: port evaluator improvements from POC2 into master

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/evaluator/evaluate_packages.go
+++ b/evaluator/evaluate_packages.go
@@ -26,13 +26,13 @@ func analyzePackages(tx *gorm.DB, system *models.SystemPlatform, vmaasData *vmaa
 		return 0, 0, errors.Wrap(err, "lazy package save failed")
 	}
 
-	pkgByName, err := loadPackages(tx, system, vmaasData)
+	pkgByName, installed, updatable, err := loadPackages(tx, system, vmaasData)
 	if err != nil {
 		evaluationCnt.WithLabelValues("error-pkg-data").Inc()
 		return 0, 0, errors.Wrap(err, "Unable to load package data")
 	}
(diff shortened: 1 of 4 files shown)


evaluator/evaluate_packages.go
```diff
@@ -137,233 +137,214 @@ func updatePackageNameDB(missing *models.PackageName) error {
...
+				EVRA:          columns.EVRA,
+				Change:        Remove,
+				ApplicableID:  columns.ApplicableID,
+				InstallableID: columns.InstallableID,
+			}
+		}
+	}
+	if err := rows.Close(); err != nil {
+		return err
 	}
```
psegedy, line 227: I think it would be better to call `defer rows.Close()` before the for loop to be sure that we released resources when ScanRows fails. To check for an error we might use `rows.Err()` after the for loop. `rows.Next()` should implicitly call `Close` once it iterates through all rows.
This seems to be the recommended way by golang docs https://go.dev/doc/database/querying
```go
defer rows.Close()
for rows.Next() {...}
if err := rows.Err(); err != nil { return err }
```

MichaelMraka: For some reason `defer rows.Close()` failed with some tests in the past. I've rechecked it and now all tests pass so I've switched back to `defer ...`.

psegedy (approved):

PR #1367, opened 2024-02-07
RHINENG-8049: improve CreateRequestRouterWithParams func

Dugowitch:
- unify `method` and `routerMethod`
- use `routerPath`, `param`, and `queryString` to crete url
- update all function calls
- update doTestWrongOffset to work with the new CreateRequestRouterWithParams func

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/manager/controllers/common_test.go
+++ b/manager/controllers/common_test.go
@@ -59,10 +59,9 @@ func CreateRequestRouterWithPath(method, routerPath, param, queryString string, 
 }
 
 // Create request and initialize router with account
-func CreateRequestRouterWithAccount(method string, url string, body io.Reader, contentType string,
-	handler gin.HandlerFunc, routerPath string, routerAccount int,
-	contextKVs ...core.ContextKV) (w *httptest.ResponseRecorder) {
-	w, req := prepareRequest(method, url, body, contentType)
+func CreateRequestRouterWithAccount(method, routerPath, param, queryString string, body io.Reader, contentType string,
+	handler gin.HandlerFunc, routerAccount int, contextKVs ...core.ContextKV) (w *httptest.ResponseRecorder) {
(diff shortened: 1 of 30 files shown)


manager/controllers/common_test.go
```diff
@@ -36,11 +37,16 @@ func CreateRequest(method string, url string, body io.Reader, contentType string
...
 
 // Create request and initialize router with params
-func CreateRequestRouterWithParams(method string, url string, body io.Reader, contentType string,
-	handler gin.HandlerFunc, routerAccount int, routerMethod string, routerPath string,
+func CreateRequestRouterWithParams(method string, routerPath string, param string, queryString string,
+	body io.Reader, contentType string, handler gin.HandlerFunc, routerAccount int,
 	contextKVs ...core.ContextKV) (w *httptest.ResponseRecorder) {
+	if len(queryString) > 0 && queryString[0] == '/' {
+		queryString = queryString[1:]
+	}
```
MichaelMraka, line 45: I think this `if` actually workarounds wrong input data. Query string should not start with '/' so it would be better to fix input data and remove the `if`.
I see there are only 2 occurrences of  "/?offset=1000" which should be "?offset=1000"
advisories_test.go:302
packages_test.go:86

MichaelMraka (changes requested):
Now I even noticed we have the same pattern also in `CreateRequestRouterWithPath()` and `CreateRequestRouterWithAccount()` functions.
Could you also fix these?
Please do 3 commits - one for each function fix.

MichaelMraka (approved):

PR #1390, opened 2024-03-05
RHINENG-8716: fix dynamically set limit in c.Request.URL

psegedy:
`c.Query(key)` claims to be a shortcut for `c.Request.URL.Query().Get(key)` by `gin` documentation. The reality is that `c.Query()` tries to get the value from QueryCache if cache exists, The cache is created with first c.Query() call on given context (might be created with other calls as well).

The problem is with /views apis. The query is added dynamically to `c.Request.URL.RawQuery` there. My guess is that it worked previously because we modified the context `c` with `apiver`... it looks like it stopped working after removal of `apiver` decisions and removal of v2 structs (it works in prod but not in stage) but I haven't tried to trace the exact commit.

IMO it should be safer to use `c.Request.URL.Query().Get(key)`

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/utils/gin.go
+++ b/base/utils/gin.go
@@ -26,7 +26,10 @@ const (
 func LoadParamInt(c *gin.Context, param string, defaultValue int, query bool) (int, error) {
 	var valueStr string
 	if query {
-		valueStr = c.Query(param)
+		// don't use c.Query(), it caches query parameters in context
+		// and data changed directly in c.Request.URL won't be returned
+		// as they are not present in the cache
+		valueStr = c.Request.URL.Query().Get(param)
 	} else {
 		valueStr = c.Param(param)
 	}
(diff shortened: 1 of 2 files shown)


MichaelMraka (approved):

psegedy:
quay is acting up, this is the error from the ci
```
15:14:16 ++ podman push quay.io/cloudservices/patchman-engine-app:pr-1390-47c7f51
15:14:17 Getting image source signatures
15:14:17 Error: trying to reuse blob sha256:0427f9bc962101ecc89f99131f22d62365ee3ddc45b57ff7b27754d500ce4ecc at destination: can't talk to a V1 container registry
```
merging it since only a comment is added with the last force push

PR #1394, opened 2024-03-07
RHINENG-7099: fix evaluator-recalc nil pointer dereference

Dugowitch:
## Why the error occurred 
The SIGSEGV originates in function `HTTPCallRetry` at line 40 (file *base/utils/http.go*):
```go
resp.Body.Close()
```
The `resp` comes from `httpCallFun` that is passed into the function as a parameter:
```go
func HTTPCallRetry(ctx context.Context, httpCallFun func() (outputDataPtr interface{},
	resp *http.Response, err error), exponentialRetry bool, maxRetries int, codesToRetry ...int) (
	outputDataPtr interface{}, err error) { ... }
```
```go
outDataPtr, resp, callErr := httpCallFun()
```
`HTTPCallRetry` function is called in `callVMaas` (file *evaluator/evaluate.go*) and `vmaasCallFunc` is passed in as `httpCallFun`.
```go
vmaasDataPtr, err := utils.HTTPCallRetry(base.Context, vmaasCallFunc, vmaasCallUseExpRetry,
	vmaasCallMaxRetries, http.StatusServiceUnavailable)
```
```go
vmaasCallFunc := func() (interface{}, *http.Response, error) { ... }
```
In `vmaasCallFunc`, `resp` comes from a `Request` method (file *base/api/client.go*). 
```go
resp, err := vmaasClient.Request(&ctx, http.MethodPost, vmaasUpdatesURL, request, &vmaasData)
```
```go
func (o *Client) Request(ctx *context.Context, method, url string,
	requestPtr interface{}, responseOutPtr interface{}) (*http.Response, error) { ... }
```
In the request method, when the request fails, the method returns *nil, err*. 
```go
httpReq, err := http.NewRequestWithContext(*ctx, method, url, body)
if err != nil {
	return nil, errors.Wrap(err, "Request failed")
}
```
Therefore `resp` in `vmaasCallFunc` as well as `resp` in `HTTPCallRetry` can be nil. A nil check is missing in `HTTPCallRetry`.

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/base/utils/http.go
+++ b/base/utils/http.go
@@ -37,7 +37,9 @@ func HTTPCallRetry(ctx context.Context, httpCallFun func() (outputDataPtr interf
 		}
 
 		responseDetails := tryGetResponseDetails(resp)
-		resp.Body.Close()
+		if resp != nil {
+			resp.Body.Close()
+		}
 		return nil, errors.Wrap(callErr, "HTTP call failed"+responseDetails)
 	}
 	return nil, errors.Errorf("HTTP retry call failed, attempts: %d", attempt)


psegedy:
Great job!
Looking at the code it seems like `nil` is only returned `Request` when we fail to make request not when we receive a http error code as I initially though. We might have incorrect data in the DB and see some other issues like pod restarts when this gets to stage, we will need to monitor it.

psegedy (approved):

PR #1442, opened 2024-06-20
RHINENG-10827: unassign systems before template delete

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/listener/templates.go
+++ b/listener/templates.go
@@ -11,10 +11,7 @@ import (
 	"strings"
 	"time"
 
-	errors2 "errors"
-
 	"github.com/pkg/errors"
-	"gorm.io/gorm"
 )
 
 const (
@@ -31,7 +28,6 @@ func TemplatesMessageHandler(m mqueue.KafkaMessage) error {
 		return nil
 	}
 
-	errs := []error{}
 	for _, template := range event.Data {
 		if enableBypass {
 			utils.LogInfo("template", template.UUID, "Processing bypassed")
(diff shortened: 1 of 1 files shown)


listener/templates.go
```diff
@@ -61,14 +61,45 @@ func TemplateDelete(template mqueue.TemplateResponse) error {
...
+		// use Find() not First() otherwise it returns error "no rows found" if uuid is not present
+		Find(&templateID).Error
+	if err != nil {
+		utils.LogWarn("template", template.UUID, "err", err)
+		return nil
+	}
+
+	// unassign systems from the template
+	err = database.DB.Model(&models.SystemPlatform{}).
+		Where("rh_account_id = ? AND template_id = ?", accountID, templateID).
```
psegedy, line 84: when the template is not found then I think it won't be error from the query above but templateID=0 would be added to this query. Could that be a problem? I hope not

psegedy (approved):
it seems like error is always nil in this function if I didn't miss something so we might not need to return error at all, otherwise LGTM

listener/templates.go
MichaelMraka: I added check for` templateID == 0`

PR #1475, opened 2024-09-06
RHINENG-10032: implement pg_repack job and admin API

Dugowitch:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [x] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/database_admin/migrations/129_create_pg_repack.up.sql
+++ b/database_admin/migrations/129_create_pg_repack.up.sql
@@ -1,0 +1,7 @@
+DO $$
+BEGIN
+    IF EXISTS (SELECT 1 FROM pg_available_extensions WHERE name = 'pg_repack') THEN
+        CREATE EXTENSION IF NOT EXISTS pg_repack;
+    END IF;
+END
+$$;
--- a/manager/routes/routes.go
+++ b/manager/routes/routes.go
@@ -113,6 +113,7 @@ func InitAdmin(app *gin.Engine, enableTurnpikeAuth bool) {
 	api.GET("/sessions", admin.GetActiveSessionsHandler)
 	api.GET("/sessions/:search", admin.GetActiveSessionsHandler)
 	api.DELETE("/sessions/:pid", admin.TerminateSessionHandler)
+	api.GET("/repack/:table_name", admin.RepackHandler)
 
 	pprof := api.Group("/pprof")
 	pprof.GET("/evaluator_upload/:param", admin.GetEvaluatorUploadPprof)
--- a/tasks/repack/repack.go
+++ b/tasks/repack/repack.go
@@ -1,0 +1,40 @@
+package repack
+
+import (
+	"app/base/core"
+	"app/base/utils"
+	"app/tasks"
+	"fmt"
+)
+
+var pgRepackArgs = []string{
+	"pg_repack", "--no-superuser-check",
+	"-d", utils.CoreCfg.DBName,
+	"-h", utils.CoreCfg.DBHost,
+	"-p", fmt.Sprintf("%d", utils.CoreCfg.DBPort),
+	"-U", utils.CoreCfg.DBUser,
+}
+
+func configure() {
+	core.ConfigureApp()
+}
+
+func Repack(table string) error {
+	// TODO
+	return nil
+}
+
+func RunRepack() {
+	tasks.HandleContextCancel(tasks.WaitAndExit)
+	utils.LogInfo("Starting repack job")
+	configure()
+
+	var table_name string = "system_package2" // FIXME: use env variable maybe
+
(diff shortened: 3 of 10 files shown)


database_admin/migrations/129_create_pg_repack.up.sql
```diff
@@ -0,0 +1 @@
+CREATE EXTENSION IF NOT EXISTS pg_repack;
```
psegedy, line 1: DB migration fails, pg_repack extension is not in the DB image used in ephemeral env.
```
Error upgrading the database: migration failed: extension "pg_repack" is not available, Could not open extension control file "/usr/share/pgsql/extension/pg_repack.control": No such file or directory. in line 0: CREATE EXTENSION IF NOT EXISTS pg_repack;
```
We could change the schema migration to try to find the extension in https://www.postgresql.org/docs/current/view-pg-available-extensions.html and run CREATE EXTENSION... only if the pg_repack extension is available

tasks/repack/repack.go
```diff
@@ -5,10 +5,13 @@ import (
...
 	"app/tasks"
 	"fmt"
+	"os"
+	"os/exec"
 )
 
 var pgRepackArgs = []string{
-	"pg_repack", "--no-superuser-check",
+	"--no-superuser-check",
+	"--no-password",
```
psegedy, line 14: why do we want to use `--no-password`? I see that password is used later in getCmd func

tasks/repack/repack.go
```diff
@@ -19,22 +22,59 @@ func configure() {
...
+// Repack runs pg_repack reindex with table. If columns are provided, cluster by these columns is executed as well.
+// Table must be a partitioned table.
+func Repack(table string, columns string) error {
+	reindexCmd := getCmd(table, "-x")
+	err := reindexCmd.Run()
+	if err != nil {
+		return err
+	}
+
+	if columns == "" || len(columns) == 0 {
```
psegedy, line 47: if len(columns) == 0 should be enough since len("") = 0

manager/routes/routes.go
```diff
@@ -113,7 +113,7 @@ func InitAdmin(app *gin.Engine, enableTurnpikeAuth bool) {
 	api.GET("/sessions", admin.GetActiveSessionsHandler)
 	api.GET("/sessions/:search", admin.GetActiveSessionsHandler)
 	api.DELETE("/sessions/:pid", admin.TerminateSessionHandler)
-	api.GET("/repack/:table_name", admin.RepackHandler)
+	api.GET("/repack/:table_name/*columns", admin.RepackHandler)
```
psegedy, line 116: shouldn't this be `/repack/:table_name/:columns` since `columns` is required according to spec? If I understand this correctly, `*columns` acts as optional parameter so the handler will also match `/repack/table_name/` and `/repack/table_name` which we actually don't need

psegedy (changes requested):

tasks/repack/repack.go
Dugowitch: It just disables prompting for a password. If a cli prompt is the only way to get a password (no env variable or .pgpass file), it will fail straight away without prompting unnecessarily

tasks/repack/repack.go
Dugowitch: Yeah, IDK how this got in there 😅

manager/routes/routes.go
Dugowitch: I believe `/repack/:table_name/*columns` is correct. OpenAPI does not allow optional path params ([OpenAPI-Specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#fixed-fields-7)), so the spec says `columns` are required even though they are not. Additionally, when tested locally, only `/repack/table_name/` and `/repack/table_name/columns` were valid and returned 200, `/repack/table_name` did not.

database_admin/migrations/129_create_pg_repack.up.sql
Dugowitch: Fixed

manager/routes/routes.go
Dugowitch: An alternative would be to implement it as a query param, or to make the param required and change the implementation accordingly. So, can it stay like this or should I change it somehow?

Dugowitch: An alternative would be to implement it as a query param, or to make the param required and change the implementation accordingly. So, can it stay like this or should I change it somehow?

psegedy (approved):
LGTM

PR #1555, opened 2025-01-15
RHINENG-15327: add templates metrics to floorist

MichaelMraka:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -521,6 +521,34 @@ objects:
           GROUP BY ra.name, am.name, at.name
           ORDER BY ra.name ASC, applicable_systems DESC;
 
+- apiVersion: metrics.console.redhat.com/v1alpha1
+  kind: FloorPlan
+  metadata:
+    name: patchman-hms
+  spec:
+    database:
+      secretName: ${FLOORIST_DB_SECRET_NAME}
+    objectStore:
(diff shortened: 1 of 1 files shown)


deploy/clowdapp.yaml
```diff
@@ -521,6 +521,34 @@ objects:
           GROUP BY ra.name, am.name, at.name
           ORDER BY ra.name ASC, applicable_systems DESC;
 
+- apiVersion: metrics.console.redhat.com/v1alpha1
+  kind: FloorPlan
+  metadata:
+    name: patchman
```
psegedy, line 527: we have a floorplan with the same `metadata.name`, I'm afraid that it wouldn't be possible to create 2 floorplans with the same name since the current one creates a CronJob with the name `floorist-patchman-exporter` and it seems like the new one would also try to create a CronJob with the same name.

@vkrizan @RoamingNoMaD is this the correct way how to define 2 FloorPlans? We need a second floorplan to push data to a different bucket. Would it be enough to provide a different `metadata.name`?

MichaelMraka: You're right. I've renamed new plan.

psegedy (approved):

PR #2156, opened 2026-04-15
RHINENG-24787: remove references to system platform

TenSt:
This PR:
- updates services to persist and query host data on the physical system_inventory and system_patch tables (joined where a full “system” row is needed) 
- introduces SystemPlatformV2 as the in-memory aggregate (Inventory + Patch)
- removes the monolithic SystemPlatform model
- adds a dedicated JOIN loader (evaluator/system_platform_v2_load.go + tests)
- threads the new types through evaluation, listener upload, manager APIs, tasks, notifications, remediations, and test helpers
- updates VMaaS JSON parsing to take SystemInventory
- updates "Deploy / FloorPlan SQL" in deploy/clowdapp.yaml to use system_inventory and system_patch
- updates Grafana panel queries to track DB size for system_inventory_* and system_patch_* partitions
- updates Docs (docs/md/architecture.md, docs/md/database.md) to describe inventory vs patch and the optional VIEW.


evaluator/evaluate.go
```diff
@@ -608,14 +624,45 @@ func updateSystemPlatform(tx *gorm.DB, system *models.SystemPlatform,
 		data["third_party"] = system.ThirdParty
 	}
 
-	err := tx.Model(system).Updates(data).Error
+	for k := range data {
+		if _, ok := systemPatchEvaluatorUpdateColumns[k]; !ok {
+			return errors.Errorf("updateSystemPlatform: unknown system_patch column %q", k)
+		}
+	}
```
MichaelMraka, line 616: This check seems to be useless.  It check that you did not set anything else than hardcoded keys from couple of lines above. How would that happened?

manager/controllers/template_systems_update.go
```diff
@@ -139,12 +139,13 @@ func assignTemplateSystems(c *gin.Context, db *gorm.DB, accountID int, template
...
 
-	// TODO: once SystemPlatform model is removed, revoke unnecessary manager privileges to system_inventory and
-	// system_patch tables. They were granted in migration 145 as a temporary fix to solve system_platform view instead
-	// of update trigger issues. Hint: have a look at 145_update_manager_privileges.down.sql
-	tx = tx.Model(models.SystemPlatform{}).
-		Where("rh_account_id = ? AND inventory_id IN (?::uuid)",
-			accountID, inventoryIDs).
+	// TODO: Revisit migration 145 manager privileges on system_inventory/system_patch; they were tied to
+	// system_platform view/update-trigger work and may be narrower now that template_id is updated on system_patch.
+	siSub := tx.Session(&gorm.Session{}).Model(&models.SystemInventory{}).
```
MichaelMraka: What's the purpose of Session() here? The query is put as subquery into next one so independent session will hardly apply.

base/database/testing.go
```diff
@@ -412,9 +413,10 @@ func DeleteTemplate(t *testing.T, account int, templateUUID string) {
 
 func CheckTemplateSystems(t *testing.T, account int, templateUUID string, inventoryIDs []string) {
 	var dbInventoryIDs []string
-	err := DB.Table("system_platform as sp").Select("sp.inventory_id as id").
-		Joins("JOIN template tp ON tp.id = sp.template_id").
-		Where("sp.rh_account_id = ? AND tp.uuid = ?::uuid", account, templateUUID).
+	err := DB.Table("system_inventory si").Select("si.inventory_id as id").
+		Joins("JOIN system_patch spatch ON si.id = spatch.system_id AND si.rh_account_id = spatch.rh_account_id").
```
MichaelMraka, line 428: I think this can go directly to `system_patch` omitting `system_inventory` completely.

tasks/caches/refresh_packages_caches.go
```diff
@@ -77,22 +77,23 @@ func getCounts(pkgSysCounts *[]models.PackageAccountData, accID *int) error {
...
 				count(*) as systems_installed,
 				count(*) filter (where spkg.installable_id is not null) as systems_installable,
 				count(*) filter (where spkg.installable_id is not null or spkg.applicable_id is not null) as systems_applicable
 			`).
-			Joins("JOIN system_package2 spkg ON sp.id = spkg.system_id AND sp.rh_account_id = spkg.rh_account_id").
-			Joins("JOIN rh_account acc ON sp.rh_account_id = acc.id").
-			Where("sp.packages_installed > 0 AND sp.stale = FALSE").
-			Group("sp.rh_account_id, spkg.name_id").
-			Order("sp.rh_account_id, spkg.name_id")
+			Joins("JOIN system_patch sp ON si.id = sp.system_id AND si.rh_account_id = sp.rh_account_id").
```
MichaelMraka, line 88: ~I don't see `system_patch` or it's columns used anywhere in the query, IMHO we can safely omit it.~
Taking back. I see `sp.packages_installed > 0`.

tasks/vmaas_sync/send_messages.go
```diff
@@ -38,9 +38,10 @@ func SendReevaluationMessages() error {
 
 func getAllInventoryIDs() ([]mqueue.EvalData, error) {
 	var inventoryAIDs []mqueue.EvalData
-	err := tasks.CancelableDB().Table("system_platform sp").
-		Select("sp.inventory_id, sp.rh_account_id, ra.org_id").
-		Joins("JOIN rh_account ra on ra.id = sp.rh_account_id").
+	err := tasks.CancelableDB().Table("system_inventory si").
+		Select("si.inventory_id, si.rh_account_id, ra.org_id").
+		Joins("JOIN system_patch sp ON si.id = sp.system_id AND si.rh_account_id = sp.rh_account_id").
```
MichaelMraka: `system_patch` or it's columns is not used anywhere in the query, IMHO we can safely omit it.

deploy/clowdapp.yaml
```diff
@@ -566,7 +566,8 @@ objects:
                  adv.applicable_systems AS applicable_systems
             FROM (SELECT sa.rh_account_id, sa.advisory_id, COUNT(*) as applicable_systems
                     FROM system_advisories sa
-                    JOIN system_platform sp ON sa.rh_account_id = sp.rh_account_id AND sa.system_id = sp.id
+                    JOIN system_inventory si ON sa.rh_account_id = si.rh_account_id AND sa.system_id = si.id
+                    JOIN system_patch spatch ON spatch.system_id = si.id AND spatch.rh_account_id = si.rh_account_id
```
MichaelMraka: Again `system_patch` not used in the rest of query, can be omitted.

evaluator/evaluate_test.go
```diff
@@ -316,20 +316,21 @@ func TestSatelliteSystemAdvisories(t *testing.T) {
...
+			YumUpdates:       yumUpdatesRaw,
+			DisplayName:      "satellite_system_test1",
+			RhAccountID:      1,
+			BuiltPkgcache:    true,
+			SatelliteManaged: true,
+		},
+		Patch: models.SystemPatch{},
 	}
-	tx := database.DB.Create(&system)
-	assert.Nil(t, tx.Error)
```
MichaelMraka, line 330: This does not feel correct, the test does not create test system anymore.

evaluator/evaluate.go
TenSt: I left it more for visibility, to have a defined list of items in one place. Removing now.

manager/controllers/template_systems_update.go
TenSt: This was meant as a rail-guard to make sure that subquery is isolated from the main query, but now looking at the code again, I see that there is no value in having separate Session here. Dropping it.

base/database/testing.go
TenSt: The test compares the list of received inventory IDs with the same list from DB. Inventory IDs are stored in `system_inventory` and template IDs are in `system_patch`, so I don't see how we can omit `system_inventory`. Am I missing something?

tasks/vmaas_sync/send_messages.go
TenSt: Yeah, I meant it as a filter (get only rows that exists in both tables), but after reviewing it again and the flow of upload - we should not have this case, so `system_patch` is redundant here and in other cases. Removing it.

base/database/testing.go
MichaelMraka: Oh I see, `system_inventory` is needed for `inventory_id`. My bad.

evaluator/evaluate_test.go
TenSt: This test doesn't need a system in DB. The `getUpdatesData` works only with the system that is passed as parameter + VMaaS cache. I don't see any DB calls, so the system doesn't need to be created in DB.

MichaelMraka: ok

MichaelMraka (approved):

PR #2157, opened 2026-04-16
RHINENG-25870: fix missing data in system_inventory

TenSt:
This PR:
- updates the list of columns that should be updated to include the latest additions for the system_inventory
- updates tests to check system for all fields that should be updated

## Summary by Sourcery

Ensure system inventory profile fields are refreshed when updating an existing system platform record.

Bug Fixes:
- Fix system_inventory updates so host-derived profile columns are refreshed on ON CONFLICT DO UPDATE, not only on first insert.

Enhancements:
- Add coverage helper to compare system_inventory profile fields against host data for consistency.

Tests:
- Extend upload and platform update tests to verify all relevant system_inventory profile fields are correctly refreshed on subsequent updates.


xbhouse:
this looks good to me! though another pair of eyes would be better :eyes: :sweat_smile:

listener/upload.go
MichaelMraka: You have to fill the model with data otherwise there's nothing to save ;).

TenSt: You mean all the fields that was added to the `colsToUpdate`? They are populated in `storeOrUpdateSysPlatform` - lines 459-473. It was done like this because we had old SystemPlatform struct and I didn't move the data population. Let me refactor it and move it to the `updateSystemPlatform` - this way we will have the data population in one place instead of two.

MichaelMraka: I mean `storeOrUpdateSysPlatform()` on L399 updates row from `systemPlatform` . 
`systemPlatform` is created on L348 and does not contain correct `OsName` etc. so it can't update it.

TenSt: Yes, but `storeOrUpdateSysPlatform` populates these data from the `host` event and only after that it passes it to `Create` in DB. I've refactored right now and moved all the data population into one place - "updateSystemPlatform" function during the init of the struct - this has to clear the confusion. Please review.

MichaelMraka (approved):

listener/upload.go
MichaelMraka: Oh, I see. Now I understand. 👍🏼

PR #2170, opened 2026-04-23
Add metric and alert for expiring certs

swadeley:
HMS-10103: add metric and alert for expiring certs in patch and update content-sources SOP

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

## Summary by Sourcery

Add certificate expiry monitoring for client certificates and wire it into the metrics subsystem.

New Features:
- Expose a Prometheus gauge reporting days until expiry for monitored client certificates, currently for the Candlepin client cert.
- Introduce a background collector that periodically refreshes certificate expiry metrics based on configured certificate/key material.
- Add a PrometheusRule manifest to alert when monitored client certificates are within 30 days of expiry.

Enhancements:
- Add a shared utility to compute days until expiration from a TLS certificate chain for reuse across components.

Documentation:
- Document the new certificate expiry alerting behavior in the PrometheusRule manifest annotations.

Tests:
- Add unit tests for the certificate expiry metric updater, including handling of invalid and missing certificate configuration.
- Add unit tests for the certificate expiry utility to cover normal, nil, unparsable, and empty certificate chains.

### DIFF
--- a/base/certutil/cert_expiry.go
+++ b/base/certutil/cert_expiry.go
@@ -1,0 +1,43 @@
+package certutil
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"time"
+)
+
+// ErrNoParseableCertificate is returned when the chain is non-empty but no DER entry could be parsed.
+var ErrNoParseableCertificate = errors.New("certutil: no parseable certificate in chain")
+
+// DaysTillExpiration returns whole days until the earliest NotAfter in the TLS certificate chain.
+// If every raw certificate fails to parse, it returns an error so callers do not treat a broken
+// configuration as “0 days left”.
+func DaysTillExpiration(certs *tls.Certificate) (int, error) {
+	expires := time.Time{}.UTC()
+	found := false
+	if certs == nil {
+		return 0, nil
+	}
+	var lastParseErr error
+	for _, tlsCert := range certs.Certificate {
+		parsed, err := x509.ParseCertificate(tlsCert)
+		if err != nil {
+			lastParseErr = err
+			continue
+		}
+		if !found || parsed.NotAfter.Before(expires) {
+			expires = parsed.NotAfter
+			found = true
+		}
+	}
+	if !found {
+		if lastParseErr != nil {
+			return 0, fmt.Errorf("certutil: parse certificate: %w", lastParseErr)
+		}
+		return 0, ErrNoParseableCertificate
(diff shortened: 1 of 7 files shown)


dev-vektr-ai:
Thanks for the clear metric/alert addition and the focused test coverage. I found two items worth addressing before this moves out of draft:

1. `DaysTillExpiration` silently ignores certificate parse failures and returns `0, nil` when every entry in the chain is unparsable. That can make a broken or malformed cert look like an already-expired/0-day cert instead of surfacing a configuration problem, which may create confusing alerts and hide the real error source.

2. `updateCertificateExpiryGauges()` returns early on `X509KeyPair` / `DaysTillExpiration` errors without updating or clearing the gauge. If the gauge was previously set, Prometheus can keep serving a stale last-good value indefinitely. Consider explicitly resetting/deleting the `candlepin` label (or setting a separate error metric) when the cert/key cannot be parsed so operators do not get a misleading expiry value.

Otherwise the collector wiring and the alert manifest look straightforward.

base/metrics/metrics.go
```diff
@@ -39,4 +43,7 @@ func Configure() {
 		mqueue.SetKafkaErrorReadCnt(KafkaConnectionErrorCnt.WithLabelValues("read"))
 		mqueue.SetKafkaErrorWriteCnt(KafkaConnectionErrorCnt.WithLabelValues("write"))
 	}
+	startCertificateExpiryCollector.Do(func() {
+		go runCertificateExpiryCollector(base.Context)
+	})
```
MichaelMraka: This is not a good place to check this metrics. `Configure()` runs in setup of every pod and I think we don't everyone to report expiration just of them will be enough :) . IMHO best place is `vmaas_sync`:  https://github.com/RedHatInsights/patchman-engine/blob/master/tasks/vmaas_sync/metrics.go#L115.
Also there's no need to run it in a loop via go routine.

MichaelMraka (changes requested):

base/metrics/metrics.go
swadeley: Hi @MichaelMraka , thank you for the review. Changes made.

MichaelMraka (approved):

PR #2176, opened 2026-04-28
RHINENG-25147: refactor workspaces

Dugowitch:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

## Summary by Sourcery

Refactor workspace handling across the service to replace JSONB-based group/workspace storage with explicit workspace_id and workspace_name columns, and propagate the new workspace-based filtering through queries, middleware, models, and exports.

Enhancements:
- Simplify workspace representation in system_inventory by introducing dedicated workspace_id and workspace_name columns and indexing them.
- Update database queries, controllers, and middleware to use workspace ID lists instead of inventory group JSON for scoping systems, advisories, packages, and templates.
- Adjust system upload and model mapping to persist a single validated workspace per host and log anomalies like multiple workspaces.
- Extend API response and export payloads to include explicit workspace identifiers and names alongside existing group data.
- Revise tests and test fixtures to align with the new workspace schema and selection semantics, including updated test data and helper context setup.

Build:
- Bump schema migration version and add a migration that backfills workspace_id and workspace_name from existing JSONB workspaces before dropping the old column.


database_admin/migrations/153_simplify_workspaces.up.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE system_inventory
+    ADD COLUMN workspace_id TEXT CHECK (NOT empty(workspace_id)),
```
MichaelMraka: workspace_id is uuid so we should use UUID type here - 2 reasons (at least):
* you won't need `empty()` check,
* and mostly UUID takes 16 bytes, while as text it takes 36 bytes .

manager/controllers/common_attributes.go
```diff
@@ -28,6 +28,11 @@ type SystemGroups struct {
 	Groups SystemGroupsList `json:"groups" csv:"groups" query:"si.workspaces" gorm:"column:groups" order_query:"si.workspaces->0->>'name'"`
 }
 
+type SystemWorkspace struct {
+	WorkspaceID   *string `json:"workspace_id" csv:"workspace_id" query:"si.workspace_id" gorm:"column:workspace_id"`
```
MichaelMraka: *uuid.UUID

base/models/models.go
```diff
@@ -74,6 +74,8 @@ type SystemInventory struct {
 	Tags                             []byte            `gorm:"column:tags"`
 	Created                          time.Time         // set by trigger system_platform_insert_trigger
 	Workspaces                       *inventory.Groups `gorm:"column:workspaces"`
+	WorkspaceID                      *string           `gorm:"column:workspace_id"`
```
MichaelMraka: *uuid.UUID

manager/controllers/advisory_systems.go
```diff
@@ -40,7 +40,6 @@ type AdvisorySystemItemAttributes struct {
 	OSAttributes
 	SystemTimestamps
 	SystemTags
-	SystemGroups
```
MichaelMraka, line 43: Removal of SystemGroups is actually trickier...
Since group column is part of API removal will break API v3 (contract). So we have 2 options:
* keep groups there as "virtual" column made from workspace_id and workspace_name,
* or create new API v4 without this column.

MichaelMraka (changes requested):

manager/controllers/advisory_systems.go
MichaelMraka: So as a part of this PR / ticket I propose keep groups as is. And I created new epic for V4 API - we have many such columns over the time so it make sense to clean things a bit.

MichaelMraka (approved):

PR #2204, opened 2026-05-21
RHINENG-25944: add search to the tags list endpoint

dominikvagner:
## Summary
Allow `/tags` to filter returned tags by matching the search term against tag namespace, key, and value.
With this the `/tags` endpoint could be used in the frontend for tag filtering.

## Summary by Sourcery

Add search support to the system tags listing endpoint and document the new query parameter.

New Features:
- Allow filtering system tags by a search term matching tag namespace, key, or value via the /tags endpoint.

Documentation:
- Document the search query parameter for the system tags list endpoint in the API documentation.

Tests:
- Add tests covering successful tag search results and empty results when no tags match the search term.

### DIFF
--- a/manager/controllers/systemtags.go
+++ b/manager/controllers/systemtags.go
@@ -1,10 +1,11 @@
 package controllers
 
 import (
-	"app/base/database"
-	"app/base/utils"
 	"errors"
 	"net/http"
+
+	"app/base/database"
+	"app/base/utils"
 
 	"app/manager/middlewares"
 
@@ -48,6 +49,11 @@ var SystemTagsOpts = ListOpts{
 	},
 	StableSort:  "tag",
 	DefaultSort: "tag",
+	SearchFields: []string{
+		"sq.tag->>'namespace'",
+		"sq.tag->>'key'",
+		"sq.tag->>'value'",
+	},
 }
 
 // @Summary Show me systems tags applicable to this application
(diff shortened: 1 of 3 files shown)


manager/controllers/systemtags.go
```diff
@@ -48,16 +49,22 @@ var SystemTagsOpts = ListOpts{
 	},
 	StableSort:  "tag",
 	DefaultSort: "tag",
+	SearchFields: []string{
+		"sq.tag->>'namespace'",
+		"sq.tag->>'key'",
+		"sq.tag->>'value'",
```
rverdile, line 55: Just a question for my understanding, how did you determine these 3 fields should be searched?

rverdile (commented):
maybe @MichaelMraka or @TenSt can double check but this lgtm! Working locally for me

manager/controllers/systemtags_test.go
```diff
@@ -76,3 +77,38 @@ func TestSystemTagsListBadRequestOnIdtSort(t *testing.T) {
...
+func TestSystemTagsListSearch(t *testing.T) {
+	core.SetupTest(t)
+
+	w := CreateRequestRouterWithAccount("GET", "/", "", "?search=k3", nil, "", SystemTagListHandler, 1)
+
+	var output SystemTagsResponse
+	CheckResponse(t, w, http.StatusOK, &output)
+
+	assert.Equal(t, 2, len(output.Data))
+	assert.Equal(t, 2, output.Meta.TotalItems)
```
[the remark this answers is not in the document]
dominikvagner: this endpoint is sorted by default, so no need for this

manager/controllers/systemtags.go
dominikvagner: yeah, this is based on how tag searching works (or is intended to work) in Insights 😅 
the tags consist of `namespace`, `key` and `value` parts, and they should be searchable on all those parts 🔍

MichaelMraka (approved):

PR #2217, opened 2026-05-29
RHINENG-26450: replace string inventory IDs with uuid.UUID

katarinazaprazna:
## Summary

Replace string inventory IDs with `uuid.UUID` across the entire backend, starting from the listener entry point (Host model) through internal models, evaluator, manager controllers, and middleware (RBAC/Kessel)

Ticket: https://redhat.atlassian.net/browse/RHINENG-26450

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

## Summary by Sourcery

Switch inventory and system identifier handling from strings to uuid.UUID across models, message payloads, controllers, and background tasks.

Enhancements:
- Update core data models and helpers to store inventory IDs as uuid.UUID instead of strings.
- Adjust HTTP handlers, middleware, and query builders to accept, validate, and query UUID-typed inventory IDs directly.
- Align Kafka/mqueue platform events, notifications, and remediations payloads to carry UUID inventory IDs consistently.
- Refactor RBAC, workspace/group parsing, and template/system advisory views to operate on UUID IDs while preserving existing response formats via string conversion.
- Update listener, evaluator, and task logic (uploads, deletes, culling, repo-based sync) to work with UUID IDs and remove ::uuid casting from SQL where possible.
- Adapt tests and test utilities throughout the codebase to the new UUID-based inventory ID types.

## Summary by Sourcery

Switch inventory/system identifiers across backend services from strings to uuid.UUID, updating models, queries, handlers, and tests while preserving external string-based representations where required.

Enhancements:
- Store inventory IDs as uuid.UUID in core data models and propagate the type change through evaluators, listeners, controllers, notification/remediation payloads, and background tasks.
- Simplify SQL by removing explicit ::uuid casts and using typed UUID parameters in queries and helper utilities.
- Tighten validation and parsing of inventory IDs and workspace/group identifiers by using uuid.Parse and uuid.UUID-aware helpers.
- Adjust test fixtures and expectations to work with UUID-typed IDs, including sorting, comparison, and serialization logic.

### DIFF
--- a/base/models/models.go
+++ b/base/models/models.go
@@ -54,9 +54,9 @@ func (TemplateBase) TableName() string {
 }
 
 type SystemInventory struct {
-	ID                               int64  `gorm:"primaryKey"`
-	InventoryID                      string `gorm:"unique"`
-	RhAccountID                      int    `gorm:"primaryKey"`
+	ID                               int64     `gorm:"primaryKey"`
+	InventoryID                      uuid.UUID `gorm:"unique"`
+	RhAccountID                      int       `gorm:"primaryKey"`
 	VmaasJSON                        *string
 	JSONChecksum                     *string
 	LastUpdated                      *time.Time `gorm:"default:null"` // set by trigger system_inventory_set_last_updated
@@ -94,9 +94,9 @@ func (SystemInventory) TableName() string {
 	return "system_inventory"
 }
 
-func (s *SystemInventory) GetInventoryID() string {
+func (s *SystemInventory) GetInventoryID() uuid.UUID {
 	if s == nil {
-		return ""
+		return uuid.Nil
 	}
 	return s.InventoryID
 }
@@ -133,9 +133,9 @@ type SystemPlatformV2 struct {
 	Patch     SystemPatch     `gorm:"embedded"`
 }
 
-func (v *SystemPlatformV2) GetInventoryID() string {
+func (v *SystemPlatformV2) GetInventoryID() uuid.UUID {
 	if v == nil {
-		return ""
+		return uuid.Nil
 	}
 	return v.Inventory.GetInventoryID()
 }
@@ -200,7 +200,7 @@ type PackageUpdate struct {
 }
 
 type DeletedSystem struct {
-	InventoryID string
+	InventoryID uuid.UUID
 	WhenDeleted time.Time
 }
 
--- a/base/mqueue/mqueue_test.go
+++ b/base/mqueue/mqueue_test.go
@@ -7,13 +7,14 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 )
 
-const id = "99c0ffee-0000-0000-0000-0000c0ffee99"
-const someid = "99c0ffee-0000-0000-0000-0000000050de"
+var id = uuid.MustParse("99c0ffee-0000-0000-0000-0000c0ffee99")
+var someid = uuid.MustParse("99c0ffee-0000-0000-0000-0000000050de")
 
-var msg = KafkaMessage{Value: []byte(`{"id": "` + id + `", "type": "delete"}`)}
+var msg = KafkaMessage{Value: []byte(`{"id": "` + id.String() + `", "type": "delete"}`)}
 
 func TestParseEvents(t *testing.T) {
 	reached := false
--- a/listener/events.go
+++ b/listener/events.go
@@ -64,6 +64,7 @@ func EventsMessageHandler(m mqueue.KafkaMessage) error {
 
 func HandleDelete(event mqueue.PlatformEvent) error {
 	defer utils.ObserveSecondsSince(time.Now(), messageHandlingDuration.WithLabelValues(EventDelete))
+
 	// TODO: Do we need locking here ?
 	err := database.OnConflictUpdate(database.DB, "inventory_id", "when_deleted").
 		Create(models.DeletedSystem{
@@ -78,7 +79,7 @@ func HandleDelete(event mqueue.PlatformEvent) error {
 	}
 
 	// mark system as stale and let system_culling job remove it later
-	query := database.DB.Model(&models.SystemInventory{}).Where("inventory_id = ?::uuid", event.ID).
+	query := database.DB.Model(&models.SystemInventory{}).Where("inventory_id = ?", event.ID).
 		Updates(map[string]interface{}{
 			"culled_timestamp": gorm.Expr("NOW()"),
 		})
--- a/manager/controllers/advisory_systems.go
+++ b/manager/controllers/advisory_systems.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"gorm.io/gorm"
 )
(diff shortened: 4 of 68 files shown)


base/models/models.go
```diff
@@ -54,9 +54,9 @@ func (TemplateBase) TableName() string {
 }
 
 type SystemInventory struct {
-	ID                               int64  `gorm:"primaryKey"`
-	InventoryID                      string `gorm:"unique"`
-	RhAccountID                      int    `gorm:"primaryKey"`
+	ID                               int64 `gorm:"primaryKey"`
+	InventoryID                      uuid.UUID
```
TenSt: I would continue to use `gorm:"unique"` even after we change the type. While `uuid` is unique by design, we're not controlling the values here (they are coming from the inventory_hosts events). To be on the safe side and make sure that we're not corrupting our data at some point in time, it is better to make sure that this field stays unique.

manager/controllers/template_systems_update.go
```diff
@@ -302,9 +302,9 @@ func checkInventoryIDs(db *gorm.DB, accountID int, inventoryIDs []string, groups
 		}
 	}
 
-	sort.Strings(missingIDs)
-	sort.Strings(satelliteIDs)
-	sort.Strings(bootcIDs)
+	sort.Slice(missingIDs, func(i, j int) bool { return missingIDs[i].String() < missingIDs[j].String() })
```
MichaelMraka: UUID is array of bytes, it's unnecessary to convert it into string for sorting
```
-	sort.Slice(missingIDs, func(i, j int) bool { return missingIDs[i].String() < missingIDs[j].String() })
+	sort.Slice(missingIDs, func(i, j int) bool { return missingIDs[i] < missingIDs[j] })
```

base/mqueue/mqueue_test.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
-const id = "99c0ffee-0000-0000-0000-0000c0ffee99"
-const someid = "99c0ffee-0000-0000-0000-0000000050de"
+var id = uuid.MustParse("99c0ffee-0000-0000-0000-0000c0ffee99")
+var someid = uuid.MustParse("99c0ffee-0000-0000-0000-0000000050de")
 
-var msg = KafkaMessage{Value: []byte(`{"id": "` + id + `", "type": "delete"}`)}
+var msg = KafkaMessage{Value: []byte(`{"id": "99c0ffee-0000-0000-0000-0000c0ffee99", "type": "delete"}`)}
```
TenSt: This is the same as `id` var, so you can use `id.String()`.

listener/events.go
```diff
@@ -64,39 +64,42 @@ func EventsMessageHandler(m mqueue.KafkaMessage) error {
 
 func HandleDelete(event mqueue.PlatformEvent) error {
 	defer utils.ObserveSecondsSince(time.Now(), messageHandlingDuration.WithLabelValues(EventDelete))
+
+	inventoryID := event.ID
```
TenSt: Why you need a separate variable here? The usual case for a separate variable is when you want to modify the value and have original left as is. Looking down the lines, I don't see any mutations, only read operations, so is there a reason for this?

manager/controllers/advisory_systems.go
```diff
@@ -197,8 +198,9 @@ func systemsIDsStatus(c *gin.Context, systems []SystemsStatusID, meta *ListMeta)
 	ids := make([]string, len(systems))
 	data := make([]IDStatus, len(systems))
 	for i, x := range systems {
-		ids[i] = x.ID
-		data[i] = IDStatus{x.ID, x.Status}
+		id := x.ID.String()
+		ids[i] = id
+		data[i] = IDStatus{id, x.Status}
```
TenSt: Same here. In theory, separate `x` would take more resources then just calling `String()` 2 times. Also, you don't need to modify the value and the scope is singular iteration.

manager/controllers/utils.go
```diff
@@ -520,8 +520,9 @@ func systemsIDs(c *gin.Context, systems []SystemsID, meta *ListMeta) (IDsPlainRe
 	resp.IDs = make([]string, 0, len(systems))
 	resp.Data = make([]IDPlain, 0, len(systems))
 	for _, x := range systems {
-		resp.IDs = append(resp.IDs, x.ID)
-		resp.Data = append(resp.Data, IDPlain{ID: x.ID})
+		id := x.ID.String()
+		resp.IDs = append(resp.IDs, id)
+		resp.Data = append(resp.Data, IDPlain{ID: id})
```
TenSt: Same here about the `id` var.

manager/controllers/utils.go
```diff
@@ -544,8 +545,9 @@ func systemsSatelliteIDs(c *gin.Context, systems []SystemsSatelliteManagedID, me
 	ids := make([]string, len(systems))
 	data := make([]IDSatelliteManaged, len(systems))
 	for i, x := range systems {
-		ids[i] = x.ID
-		data[i] = IDSatelliteManaged{x.ID, x.SatelliteManaged}
+		id := x.ID.String()
+		ids[i] = id
+		data[i] = IDSatelliteManaged{id, x.SatelliteManaged}
```
TenSt: Same here about the `id` var.

tasks/system_culling/system_culling_test.go
```diff
@@ -213,12 +214,12 @@ func TestPruneDeletedSystems(t *testing.T) {
...
 		invID := fmt.Sprintf("00000000-0000-0000-0000-000000000de%d", i+1)
 		assert.NoError(t, database.DB.Create(&models.DeletedSystem{
-			InventoryID: invID,
+			InventoryID: uuid.MustParse(invID),
 			WhenDeleted: staleDate,
 		}).Error)
 	}
 	assert.NoError(t, database.DB.Create(&models.DeletedSystem{
-		InventoryID: "00000000-0000-0000-0000-000000000deff",
+		InventoryID: uuid.MustParse("00000000-0000-0000-0000-0000000000de"),
```
TenSt, line 222: Please double-check that the ID change is valid as new one is different from the old one.

manager/controllers/template_systems_update.go
```diff
@@ -302,9 +303,9 @@ func checkInventoryIDs(db *gorm.DB, accountID int, inventoryIDs []string, groups
 		}
 	}
 
-	sort.Strings(missingIDs)
-	sort.Strings(satelliteIDs)
-	sort.Strings(bootcIDs)
+	sort.Slice(missingIDs, func(i, j int) bool { return bytes.Compare(missingIDs[i][:], missingIDs[j][:]) < 0 })
+	sort.Slice(satelliteIDs, func(i, j int) bool { return bytes.Compare(satelliteIDs[i][:], satelliteIDs[j][:]) < 0 })
+	sort.Slice(bootcIDs, func(i, j int) bool { return bytes.Compare(bootcIDs[i][:], bootcIDs[j][:]) < 0 })
```
katarinazaprazna, line 308: @MichaelMraka, @TenSt I attempted to use the suggestion for a simpler comparison, but it kept failing. The `uuid.UUID` type is [16]byte, and Go does not seem to support < on arrays. To get rid of the string casting, I'm now comparing the underlying byte slices instead. What do you think about this approach?

TenSt: Yeah, that would work, good approach 👍

TenSt (approved):
lgtm!

PR #2235, opened 2026-06-18
RHINENG-27056: fix waitForSessionClosed query

TenSt:
This PR:

- Fixes `waitForSessionClosed` to use lib/pq `ANY($1)` and fail closed on query errors
- Replaces GORM-style `IN (?)` with `usename = ANY($1)` and `pq.Array(lockUsers)` for raw `database/sql`
- Extracts `findActiveAppSession()` so session lookup is testable and errors are handled explicitly
- Retries on query failure instead of treating errors as “no sessions found” (previous fail-open behavior)
- Adds tests for error path, no sessions, and active `manager` session detection

Test steps locally:
- `docker compose -f docker-compose.test.yml run --rm test ./scripts/go_test.sh './database_admin'` (with DB already migrated)

## Summary by Sourcery

Ensure database session checks for application lock users are robust against query errors and correctly detect active sessions.

Bug Fixes:
- Correct session lookup to use a parameterized ANY() query compatible with database/sql and lib/pq.
- Treat query failures as errors that are retried instead of incorrectly assuming no active sessions (fail-closed behavior).

Enhancements:
- Extract session lookup into a dedicated findActiveAppSession helper to improve testability and explicit error handling.

Tests:
- Add unit and integration-style tests covering query failures, absence of sessions, and detection of an active manager session.

### DIFF
--- a/database_admin/update.go
+++ b/database_admin/update.go
@@ -3,16 +3,25 @@ package database_admin
 import (
 	"app/base/utils"
 	"database/sql"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/golang-migrate/migrate/v4/database"
+	"github.com/lib/pq"
 	log "github.com/sirupsen/logrus"
 )
 
 var lockUsers = []string{"listener", "evaluator", "manager", "vmaas_sync"}
+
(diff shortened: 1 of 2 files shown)


swadeley (approved):
ACK

PR #2236, opened 2026-06-18
RHINENG-27056: move db-migration to a separate job

TenSt:
### Context

Manager had db-migration as an init container. With multiple replicas and rolling updates, several new pods could run migration at once and compete for pg_advisory_lock(123). ClowdApp cannot set maxSurge on manager (public web service enabled); a dedicated Job avoids per-replica migration. Also, it is harder to troubleshoot when you have 2 pods trying to do migration and you don't know which one actually does it.

### This PR:

- Moves DB migration from manager init container to a single ClowdApp Job to avoid concurrent migrators during rollout
- Replaces manager `db-migration` init with `check-for-db` (same pattern as listener/evaluator)
- Adds `db-migration` Job
- Updates `entrypoint.sh` to retry migrate on failure

### Summary of the new flow: 
New deploy → migration Job and new pods start in parallel → new pods block in init until DB schema is upgraded → success = rollout completes, failure = new pods fail init and old pods keep serving.

## Summary by Sourcery

Move database migration from the manager init container to a dedicated ClowdApp Job and make migrations more resilient to failures.

New Features:
- Introduce a dedicated db-migration Job to run schema migrations independently of manager pods.

Enhancements:
- Replace the manager db-migration init container with a lightweight check-for-db init container that only waits for the upgraded schema.
- Add configurable migration timeout and retry logic for the migration entrypoint to improve robustness during rollouts.

Deployment:
- Configure a MIGRATION_TIMEOUT parameter to control the db-migration Job execution time.

### DIFF
--- a/database_admin/entrypoint.sh
+++ b/database_admin/entrypoint.sh
@@ -1,8 +1,24 @@
 #!/bin/bash
 
-set -e -o pipefail # stop on error
+set -e -o pipefail
 
 MIGRATION_FILES=file://./database_admin/migrations
+MAX_RETRIES=${MIGRATION_MAX_RETRIES:-1}
+
+run_migration() {
+  ${GORUN:+go run} ./main${GORUN:+.go} migrate $MIGRATION_FILES
+}
 
 echo "Running in $(pwd) as $(id)"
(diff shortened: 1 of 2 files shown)


dominikvagner (approved):
looks good to me! ✅

PR #2246, opened 2026-06-24
RHINENG-27056: update cji db-migration job with name using image_tag and suffix

TenSt:
This PR:
- updates the CJI db-migration with name using image_tag and random suffix because `qontract-reconcile` does not support `generateName` for any k8s object

## Summary by Sourcery

Update the CJI db-migration job to use an explicit name incorporating the image tag and a generated unique suffix instead of relying on Kubernetes generateName, and add a new parameter to supply that suffix.


swadeley (approved):
ACK

PR #2247, opened 2026-06-25
docs: add small update to trigger stage deployment

TenSt:
This PR adds very small docs update to trigger stage deployment as previous one failed for the unknown reason.

## Summary by Sourcery

Documentation:
- Update README to state that local deployment can use either podman-compose or docker compose for running project components and supporting containers.


swadeley (approved):
ACK

PR #2249, opened 2026-06-25
HMS-9951: store template-advisory relationships on template update

xbhouse:
## Summary 

On template update events:
- Calls content-sources API to fetch the current advisories for a template
- Calculates the diff between returned and stored advisories
- Stores updated relationships in the template_advisory table

On template delete events:
- Removes the template_advisory entries for the deleted template

## Testing steps

1. Build and start the app: `podman-compose up --build`
2. Send a template update event to the template topic: `curl -X PUT http://localhost:9001/control/templates`. There should be a template-updated message in the logs.
3. The templates' advisories should be added to the template_advisory table: 
```
podman exec -it db psql -d patchman -U admin

patchman=# select * from template_advisory;
 rh_account_id | template_id | advisory_id 
---------------+-------------+-------------
             1 |         112 |           1
             1 |         112 |           3
             1 |         113 |           1
             1 |         113 |           3
```
4. Send the same update event. The template_advisory table should be unchanged.
5. Send a template delete event: `curl -X DELETE http://localhost:9001/control/templates`. The templates' advisories should be removed from the template_advisory table.

## Summary by Sourcery

Integrate content-sources advisory data into template processing so that template-advisory relationships are kept in sync on update and cleaned up on delete, with supporting utilities and tests.

New Features:
- Synchronize template-to-advisory relationships with content-sources on template update events.
- Expose a content-sources client and mock HTTP endpoint to retrieve advisory IDs for templates.

Enhancements:
- Track template listener metrics per event type instead of assuming create events.
- Remove template-advisory relationships when templates are deleted to keep data consistent.
- Add utilities for encoding and generating x-rh-identity headers for outbound service calls.

Tests:
- Add unit and integration tests covering advisory sync diffing, database lookups, and content-sources interactions.
- Add database test helpers for creating, checking, and deleting template-advisory records.

### DIFF
--- a/listener/template_advisories.go
+++ b/listener/template_advisories.go
@@ -1,0 +1,199 @@
+package listener
+
+import (
+	"app/base/content_sources"
+	"app/base/database"
+	"app/base/models"
+	"app/base/utils"
+	"context"
+	"net/http"
+
+	"github.com/pkg/errors"
+	"github.com/redhatinsights/platform-go-middlewares/v2/identity"
+	"gorm.io/gorm"
+)
+
+func syncTemplateAdvisories(ctx context.Context, accountID int, templateID int64, templateUUID string) error {
+	current, err := callCSTemplateAdvisories(ctx, templateUUID)
+	if err != nil {
+		return errors.Wrap(err, "fetching current template advisories from content-sources")
+	}
+
+	stored, err := lookUpTemplateAdvisories(database.DB, accountID, templateID)
+	if err != nil {
+		return errors.Wrap(err, "looking up stored template advisories")
+	}
+
+	toAdd, toRemove := diffTemplateAdvisories(current.AdvisoryIDs, stored)
+	if len(toAdd) == 0 && len(toRemove) == 0 {
+		utils.LogInfo("template_uuid", templateUUID, "template advisories unchanged")
+		return nil
+	}
+
+	nameToID, err := lookUpAdvisoryMetadataIDs(toAdd)
+	if err != nil {
+		return errors.Wrap(err, "looking up advisory metadata IDs")
+	}
+
+	toInsert := buildTemplateAdvisoryRows(accountID, templateID, toAdd, nameToID)
+
+	tx := database.DB.Begin()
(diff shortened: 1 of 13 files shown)


listener/templates.go
```diff
@@ -134,7 +142,15 @@ func TemplateUpdate(template mqueue.TemplateResponse) error {
 	err = database.OnConflictUpdateMulti(database.DB, []string{"rh_account_id", "uuid"},
 		"name", "environment_id", "description", "creator", "published").Save(&row).Error
 	if err != nil {
-		return errors.Wrap(err, "creating template from message")
+		return errors.Wrap(err, "creating/updating template from message")
+	}
+
+	if contentSourcesBaseURL != "" && eventType == TemplateEventUpdate {
+		ctx := identity.WithIdentity(context.Background(), utils.XRHIDForOrg(template.OrgID))
```
xbhouse, line 149: i'm not sure of the best approach to construct the identity for the request to content-sources. i've done it this way since the request org ID needs to match the org ID of the template, and since this request doesn't come through the UI, i don't think we can use the identity from rbac. let me know if there's an existing pattern or a better way to do this!

listener/template_advisories.go
```diff
@@ -0,0 +1,188 @@
...
+			TemplateID:  templateID,
+			AdvisoryID:  advisoryID,
+		})
+	}
+	return rows
+}
+
+// Deletes given template advisory relationships from template_advisory
+func deleteOldTemplateAdvisories(tx *gorm.DB, accountID int, templateID int64, advisoryIDs []int64) error {
+	err := tx.Where("rh_account_id = ? ", accountID).
```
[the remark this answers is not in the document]
TenSt: @xbhouse I agree with this part, the `len` check seems reasonable.

listener/templates.go
TenSt: I don't know about any other way and we can refactor later if there is a better way 👍

TenSt (approved):
lgtm! great job

PR #2252, opened 2026-06-30
RHINENG-26549: add glitchtip integration, fix logging calls

dominikvagner:
This adds integration with GlitchTip through the sentry Go SDK. It will
try to set up the client if the sentry DSN is present in the config and
then capture exceptions and messages with error, fatal or panic level.
When captured some context is added, and then it's sent as an event.

_Related to GlitchTip logging, logging util functions calls were 
changed to **not** pass stringified errors as arguments._
> _[fix: stop stringifying errors when logging]_
> Errors don't need to be stringified when being passed into log functions
> as they will be turned into strings there. Having them as proper errors
> will enable better error logging with sentry/glitchtip, stack traces
> will be present.

Associated Ticket: [[RHINENG-26549](https://redhat.atlassian.net/browse/RHINENG-26549)]

[RHINENG-26549]: https://redhat.atlassian.net/browse/RHINENG-26549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/base/base.go
+++ b/base/base.go
@@ -56,7 +56,7 @@ func TryExposeOnMetricsPort(app *gin.Engine) {
 	}
 	err := utils.RunServer(Context, app, metricsPort)
 	if err != nil {
-		utils.LogError("err", err.Error())
+		utils.LogError("err", err)
 		panic(err)
 	}
 }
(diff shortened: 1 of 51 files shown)


Dugowitch (approved):
LGTM 🚀

PR #2264, opened 2026-07-09
fix: add user for patch to call content sources API

xbhouse:
To fix the rejected requests from patch to content sources:

- Adds a user environment variable to the deployment
- Adds this user to the identity of the request to content sources endpoint that lists template advisory IDs for a specific template
- If this user is empty, no request is sent to content sources

Related to https://github.com/content-services/content-sources-backend/pull/1556

## Summary by Sourcery

Add support for configuring a content sources user and include it in identity used for content sources requests.

Enhancements:
- Extend core configuration and deployment to provide a CONTENT_SOURCES_USER environment parameter.
- Include the configured content sources user in the XRHID identity used for content sources calls.
- Guard template advisory synchronization so it only runs when a content sources user is configured.

Tests:
- Update identity tests to validate the content sources user is included in the generated XRHID.

### DIFF
--- a/base/utils/config.go
+++ b/base/utils/config.go
@@ -82,6 +82,7 @@ type coreConfig struct {
 	EvaluatorUploadPrivateAddress string
 	EvaluatorRecalcPrivateAddress string
 	ContentSourcesAddress         string
+	ContentSourcesUser            string
 
 	// cloudwatch
 	CloudWatchAccessKeyID     string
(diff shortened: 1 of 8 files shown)


xbhouse:
related MR: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/196165 

i haven't set deployment variables like this in patch before so let me know if it looks off :)

TenSt (approved):
lgtm!

